PR #7, opened 2019-12-09
Try variable image with args

webbertakken:
While waiting for https://github.com/webbertakken/unity-actions/issues/3#issuecomment-560118513 to get a reply, this PR attempts to try all possible options to get variable images through args or environment variables.


webbertakken:
Trying arg substitution as per Docker docs https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

webbertakken:
Looks like all available options have been exhausted. We need support for passing arguments on build time from the Actions team.

PR #8, opened 2019-12-14
Initial support for multiple build targets

webbertakken:
This PR will add support for building different platforms based on different action releases.

Ultimately the plan is to support all builds form the same version as discussed in https://github.com/webbertakken/unity-actions/issues/3. In the meantime we should already be able to build for all targets.

Acceptance criteria is to have all builds run in parallel. https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobs

This is a work in progress


webbertakken:
The above means that we effectively can not use matrix to determine the version of an action.

That even increases the need for being able to run all builds with different targets and versions from an action with the same version.

webbertakken:
Last few commits start into a direction of javascript running a docker container based on the input.

webbertakken:
This works. Now need some cleanup and ensuring stability before merging.

webbertakken:
Reference: Multiple build targets: https://github.com/webbertakken/unity-actions/issues/12
Reference: Multiple images: https://github.com/webbertakken/unity-actions/issues/3

webbertakken:
Reference: Multiple images: https://github.com/webbertakken/unity-actions/issues/3

PR #9, opened 2020-01-06
Fix build path

webbertakken:


webbertakken:
Continued in https://github.com/webbertakken/unity-builder/pull/10

PR #10, opened 2020-01-06
Always use relative path to action runner workspace

webbertakken:


webbertakken:
Fixes https://github.com/webbertakken/unity-actions/issues/23

PR #11, opened 2020-01-07
README now mentions allowed targetPlatform values

Rabadash8820:


Rabadash8820:
No idea what to make of those failed checks...

webbertakken:
Thanks for the PR. The checks fail because your fork does not have `UNITY_LICENSE` set in secrets. 

Currently this is necessary in order to test the actions as part of the workflow. However I'm looking for a better way to solve this.

For now if you just set the variable, it should work.

webbertakken:
Actually I'll just merge this one

PR #14, opened 2020-01-11
docker.js now sets variables for Pro licenses as well

Rabadash8820:
Not sure if anything else needs to be updated, but this should solve my Issue at webbertakken/unity-actions#26.

I set `UNITY_EMAIL`, `UNITY_PASSWORD`, and `UNITY_SERIAL` in my fork so that the test builds would pass, but my Plus license is limited to only 2 activations at a time, so the remaining builds may still fail...

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,6 +6,9 @@ on:
 
 env:
   UNITY_LICENSE: ${{ secrets.UNITY_LICENSE }}
+  UNITY_EMAIL: ${{ secrets.UNITY_EMAIL }}
+  UNITY_PASSWORD: ${{ secrets.UNITY_PASSWORD }}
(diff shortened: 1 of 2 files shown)


Rabadash8820:
Hm, looks like the `docker run` in your test builds are still missing the environment variables that I added:

image

webbertakken:
> Hm, looks like the `docker run` in your test builds are still missing the environment variables that I added:
> 
> image

Hmm I see. I don't know exactly how actions work in this regard, it looks like the tests are run on master somehow. 

Do you think I can merge it without breaking the status?

webbertakken:
Since the badge was showing red anyway, I merged it. Do you have any idea what's wrong with my workflow?

DanVicarel:
Ok cool. I'm still pretty new to the details of GitHub Actions, but if I figure somthin out I'll bring it up on webbertakken/unity-actions#27.

PR #15, opened 2020-01-12
Only return the license in professional mode

Jibbajabbafic:
If $UNITY_SERIAL isn't defined then the license return step gets skipped to help save some time in personal builds.

### DIFF
--- a/builder/steps/return_license.sh
+++ b/builder/steps/return_license.sh
@@ -1,14 +1,16 @@
 #!/usr/bin/env bash
 
-#
-# PROFESSIONAL (SERIAL) LICENSE MODE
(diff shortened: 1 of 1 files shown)


Jibbajabbafic:
Looks like it's failing because UNITY_LICENSE isn't defined?

webbertakken:
@Jibbajabbafic perfect. Thanks for your PR.

The error is related to https://github.com/webbertakken/unity-actions/issues/27

Honestly I don't know how to fix that. You have an idea?

Jibbajabbafic:
No idea, I'll try and investigate at some point when I have time

PR #16, opened 2020-01-13
Testing PR checks

Rabadash8820:
Just testing to see if I can get a PR-triggered workflow to correctly use the action defined in the PR.

Using a Draft PR, not sure if that will affect checks running...

Edit: sweet, they still do


Rabadash8820:
@webbertakken Ok so that last check run was enlightening. See the [results here](https://github.com/webbertakken/unity-builder/commit/fb5e26daa15a37c2be261d4fea92bd62010a6ab0/checks?check_suite_id=398805311). I tweaked `builder/entrypoint.sh` so it just logs something then early exits. When I pushed that commit, the workflow was triggered in this repo, and the Action output shows the log message. In other words, the workflow _did_ run the latest code present in the PR.

webbertakken:
Could you be more specific when you say this repo?

Please note that I don't have variables for Pro license defined.

Rabadash8820:
So from this last check run I think I've identified the issue that we've been looking for since I opened webbertakken/unity-actions#26. It looks like the `UNITY_*` environment variables are simply empty when the workflow runs. I tested this by logging `process.env` in the Action's `builder/index.js` file, and then logging `UNITY_EMAIL` in `activate.sh` ([not the most secure thing](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets#limits-for-secrets) to do in hindsight, but luckily nothing was printed). See the [results here](https://github.com/webbertakken/unity-builder/commit/85e6ed52f5d169849e5bf94225eab157bedbde5a/checks?check_suite_id=399248305).

Again, `UNITY_LICENSE`, `UNITY_SERIAL`, `UNITY_EMAIL`, and `UNITY_PASSWORD` are all _defined_ in the PR workflow, they're just apparently empty by the time this action runs. I have no why that would be. My fork has those secrets defined too, so even if the Action we're using the fork's secrets, they should be non-empty...

Rabadash8820:
> Could you be more specific when you say this repo?

I mean `unity-builder`...comment edited.

webbertakken:
> I mean unity-builder...comment edited.

Sorry I meant upstream or downstream. I'm assuming by your phrasing that you mean upstream. 

So how do you think we should solve this? Use the `with` keyword perhaps? 

If you don't find a solution I will have a look at it next week, when I have access to a computer again. Thanks for the effort at least @Rabadash8820

Rabadash8820:
>  I'm assuming by your phrasing that you mean upstream.

Yes, sorry, I did mean upstream.

> So how do you think we should solve this? Use the `with` keyword perhaps?

Yeah I mean once again I'm in a little over my head. I'm not sure why GitHub wouldn't be sending the `env` to your Docker images, and I don't know if/how `with` works differently. Seems worth a try though. I like the idea of using `with` regardless, so that all inputs to your Actions are in one place. If users still prefer to use `env`, then they could just set the `INPUT_*` environment variables that GitHub reads into inputs by default.

I haven't been able to replicate your dev environment locally (I haven't used Prettier, Yarn, husky, or ncc before), so I'm not super confident that I'll be able to make a PR if any JS needs to change, but I'll see what I can do.

Jibbajabbafic:
FYI this might be the issue:
```
With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.
```
From: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets#using-encrypted-secrets-in-a-workflow

Jibbajabbafic:
So after looking into it a bit more I don't think there's any way to pass a secret to run on a PR since it would be trivial for anyone to open a malicious PR and leak the secret.

Rabadash8820:
Nice catch @Jibbajabbafic, I think you're right. The Issue that kind of started all this discussion (webbertakken/unity-actions#26) is still unresolved, but based on everything we've learned from this PR and our myriad other discussions, I don't think there's any reason to keep this PR open. Closing now...

PR #17, opened 2020-01-15
feat(activate): add more checks for licenses

Jibbajabbafic:
Adds more checks for professional license details before running a build.
This should stop builds running without a licenses setup and will exit with a more explicit error

Edit: also removed the exit from the professional activation since that was being called even when activation was successful

### DIFF
--- a/builder/steps/activate.sh
+++ b/builder/steps/activate.sh
@@ -1,8 +1,6 @@
 #!/usr/bin/env bash
 
-# Try personal activation
 if [[ -n "$UNITY_LICENSE" ]]; then
-  LICENSE_MODE="personal"
   #
   # PERSONAL LICENSE MODE
   #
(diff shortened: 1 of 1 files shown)


builder/steps/activate.sh
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 
+# Try personal activation
```
webbertakken: Please stick to the already existing comments style (inside the if blocks).

builder/steps/activate.sh
```diff
@@ -1,6 +1,8 @@
...
 if [[ -n "$UNITY_LICENSE" ]]; then
+  LICENSE_MODE="personal"
```
webbertakken: I like this approach a lot! Could you place this variable under the comment though?

builder/steps/activate.sh
```diff
@@ -42,30 +44,20 @@ if [[ -n "$UNITY_LICENSE" ]]; then
...
+elif [[ -n "$UNITY_SERIAL" && -n "$UNITY_EMAIL" && -n "$UNITY_PASSWORD" ]]; then
+  LICENSE_MODE="professional"
   #
   # PROFESSIONAL (SERIAL) LICENSE MODE
   #
   # This will activate unity, using the activating process.
   #
   # Note: This is the preferred way for PROFESSIONAL LICENSES.
   #
+
```
webbertakken: Unnecessary new whitespace

PR #27, opened 2020-01-26
Add cache step to workflow

webbertakken:
Relates to https://github.com/webbertakken/unity-actions/issues/34


webbertakken:
Before cache is hit (WebGL)

image

webbertakken:
After cache is hit (WebGL) `5 minutes faster`

image

PR #66, opened 2020-03-01
Support put UNITY_LICENSE in file

litefeel:


action/steps/activate.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-if [[ -n "$UNITY_LICENSE" ]]; then
+if [[ -n "$UNITY_LICENSE" || -n "$UNITY_LICENSE_PATH" ]]; then
```
webbertakken: note that for this to be correct, the or operator (`||`) has to be outside of the square brackets.

PR #69, opened 2020-03-04
Add UNITY_LICENSE_PATH

litefeel:


action/steps/activate.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-if [[ -n "$UNITY_LICENSE" ]]; then
+if [[ -n "$UNITY_LICENSE" || -n "$UNITY_LICENSE_PATH" ]]; then
```
webbertakken: Or operator should be outside of the double square brackets for compatibility reasons

`if [[ -n "$UNITY_LICENSE" ]] || [[ -n "$UNITY_LICENSE_PATH"  ]]; then`

action/steps/activate.sh
```diff
@@ -15,8 +15,14 @@ if [[ -n "$UNITY_LICENSE" ]]; then
...
   FILE_PATH=UnityLicenseFile.ulf
 
-  # Copy license file from Github variables
-  echo "$UNITY_LICENSE" | tr -d '\r' > $FILE_PATH
+  if [[ -n "$UNITY_LICENSE" ]]; then
+    # Copy license file from Github variables
+    echo "$UNITY_LICENSE" | tr -d '\r' > $FILE_PATH
+  else
+    # Copy license file from file system
+    cat "$UNITY_LICENSE_PATH" | tr -d '\r' > $FILE_PATH
```
webbertakken: This else implies that any other case is using a file path, which is not the case.

The variable should also correctly be called `UNITY_LICENSE_FILE`, as it is a file, not a path.

litefeel:
> In general, is there a reason why you're not simply renaming your distributed license file to `UnityLicenseFile.ulf`? It should already work without any code changes.

A license can only be used by one unity version.
When running on multiple unity versions, we must have multiple license files.

action/steps/activate.sh
litefeel: image

https://linuxacademy.com/blog/linux/conditions-in-bash-scripting-if-statements/

The syntax I see from here.

webbertakken: I understand that works, but in my experience it's generally better to not have the or part of the condition at all.

There's a good plugin for modern code editors called `shellcheck` that will help you validate your script if you're afraid to make changes.

action/steps/activate.sh
webbertakken: Could you make this an `elif` still?

webbertakken:
Strange that the test is failing. I'll have a look at that.

litefeel:
Hello.
-I manually modified index.js. I try to revert it. But the CI still fails.
-I built index.js via `yarn build`, but the files look different.
Like this.

image

webbertakken:
I merged your PR. the different of the index file seemed to be caused by not updated dependencies.

To fix this next time you should `yarn` to update the dependencies (because you rebased) and then `yarn build`, instead of only doing `yarn build` again. Either way it should have gone automatically, so I will fix this in another PR here https://github.com/webbertakken/unity-builder/pull/74.

Thanks again for your contributions!

PR #80, opened 2020-04-21
Introduce build versioning

webbertakken:
This PR:
- Adds versioning strategies that make versioning work out of the box (does not apply for custom build scripts).
- Add strategy `Semantic` which will be based on tags and the number of commits on top. (0.1.2)
- Add strategy `None` to give the option to opt out. (version will be unchanged)
- Add strategy `SemanticClassic` which will only be based on the tag (tag must start with v#, where # is a digit for this to work)
- Add strategy `Custom` to simply pass in a specific version of your own choosing.

This last option would be useful for pipelines that already generate their versions based on a different script. The implementation for that would look like this (supplying `version` in addition to `versioning`:

```yaml
  versioning: Custom
  version: "0.4.12"
```

The default strategy will be set to `Semantic`, because it allows for easy out of the box versioning, which helps the newer developers under us (but also pro's) succeed faster. 🎉

This means that versioning is effectively opt-out. 😉


webbertakken:
This is still a work in progress. Feel free to provide any comments or feedback!

Probably the commands directly to git [here](https://github.com/webbertakken/unity-builder/pull/80/files#diff-10873d802b18987a9959a13c6bcfcd5e) can be abstracted into a Versioning.Generator class, which could take responsibility for calling the git commands.

webbertakken:
Most of the code seems to be working well.

The current [errors](https://github.com/webbertakken/unity-builder/pull/80/checks?check_run_id=610114434#step:4:2978) may be caused by one of these things:

- the object that was checked out is not recognising tags
- repos that are accessed from within the docker container are somehow nested

Perhaps the better approach is to already figure out the version in the javascript part in the first place and just set it during the build phase.

webbertakken:
Looks like in the second to last run it showed up properly as the branch name, but the very last one didn't. 

image

image

This looks random at first sight.

webbertakken:
Everything works from javascript now, except for the part that if someone commits to the same branch, and then someone restarts the pipeline, there may be no way to find out how many commits we're behind, due to:
image

webbertakken:
Next finding:

The sha is nothing like that of the original commit. So on a pull request there is very little you can relate to the original commit. As can be seen in one of the results here:

image

The first hash is from describe. the second one is GITHUB_SHA.

Looks like at least for pull requests the official docs are lying, or i'm missing something that's not all that intuitive.

image

PR #91, opened 2020-05-21
Add flag for allowing dirty branch

webbertakken:
This adds the flag `allowDirtyBuild` which takes a boolean-string.

When the flag is set to true, versions can be generated based on a dirty branch.

This flag was added when folders like `streamingAssets` have to be deleted for building for specific platforms.

Resolves https://github.com/webbertakken/unity-builder/issues/89


webbertakken:
This PR includes a quick refactor that splits responsibility between Input and BuildParameters to make changes like this possible:

image

BuildParameters now relies on Input and Version

image

Input is now a simple model with getters and some basic input sanitisation. Now without any dependencies outside of @actions/core.

image

PR #99, opened 2020-06-11
make StandaloneOSX app runnable

nathanleiby:
Potential fix for: https://github.com/webbertakken/unity-builder/issues/77

Questions:
- The contributor guidelines request adding tests, but given that this happens in the shell script I wasn't certain if/how to do that best. Any suggestions, or is a manual verification good enough?
- In the issue it discusses fixing Linux builds as well, but I haven't yet tested if they're broken or determined the path where the executable would need to be updated. Do we want to include that here as well or just Mac?

### DIFF
--- a/action/steps/build.sh
+++ b/action/steps/build.sh
@@ -123,6 +123,11 @@ else
   echo "Build failed, with exit code $BUILD_EXIT_CODE";
 fi
 
+# Add permissions to make app runnable
+if [[ "$BUILD_TARGET" == "StandaloneOSX" ]]; then
+  chmod +x $BUILD_PATH_FULL/StandaloneOSX.app/Contents/MacOS/*
+fi
+
 #
 # Results
 #
--- a/action/steps/build.sh
+++ b/action/steps/build.sh
@@ -125,7 +125,9 @@ fi
 
 # Add permissions to make app runnable
 if [[ "$BUILD_TARGET" == "StandaloneOSX" ]]; then
-  chmod +x $BUILD_PATH_FULL/StandaloneOSX.app/Contents/MacOS/*
+  ADD_PERMISSIONS_PATH=$BUILD_PATH_FULL/StandaloneOSX.app/Contents/MacOS/*
+  echo "Making the following path executable: $ADD_PERMISSIONS_PATH"
+  chmod +x $ADD_PERMISSIONS_PATH
 fi
 
 #


nathanleiby:
After adding logging, I now see that it ran here (https://github.com/webbertakken/unity-builder/runs/760264818):

```
Making the following path executable: /github/workspace/build/StandaloneOSX/StandaloneOSX.app/Contents/MacOS/*
```

action/steps/build.sh
```diff
@@ -123,6 +123,13 @@ else
   echo "Build failed, with exit code $BUILD_EXIT_CODE";
 fi
 
+# Add permissions to make app runnable
+if [[ "$BUILD_TARGET" == "StandaloneOSX" ]]; then
+  ADD_PERMISSIONS_PATH=$BUILD_PATH_FULL/StandaloneOSX.app/Contents/MacOS/*
```
webbertakken, line 128: Do all contents of that folder have to be given the executable bit?

nathanleiby: So far, I've only seen a single file within that directory, i.e.
```
$BUILD_PATH_FULL/StandaloneOSX.app/Contents/MacOS/<my project name>
```

I used the chmod command with `*` based on recommended command in the issue. @dyarosla what was your experience? Thanks!

webbertakken: Ok fair enough.

Sorry to not bring this up in the initial review but `StandaloneOSX` comes from targetPlatform if i'm not mistaken and .app is added by unity itself?

nathanleiby:
I pushed a commit to use this build step (https://github.com/CleverGameDev/Cleverfall/commit/527980246bf136558fd7beb9226ecbd0ee459b33) and verified that the output is now executable (https://clevergamedev.itch.io/cleverfall). 

NOTE: You still have do a security override to allow running a binary from unverified developer, but after that it works as expected vs hitting a 2nd error. This was expected, it's just standard MacOS security.

Thanks for shepherding this out! 🐑

mastef:
@nathanleiby @webbertakken so I recently tried the MacOS build, and the resulting `StandaloneOSX.app` and the `StandaloneOSX.app/Contents/MacOS/ProgramName` were not executable. So double-clicking on the `StandaloneOSX.app` would come up with an error :
image

Trying to run it through terminal, would throw this error : 
```
$ open StandaloneOSX.app/
LSOpenURLsWithRole() failed with error -10810 for the file /Downloads/Build/StandaloneOSX/StandaloneOSX.app.
```

After running `chmod +x` on both of them, I was able to open the app.

Maybe a regression is happening? Or the `chmod +x` is getting lost in the .zip?

webbertakken:
@mastef thank you for reporting your issue.

Could you please open a new issue for this that is not closed and copy paste all information you've just provided?

That way also other contributors may be able to help.

mastef:
@webbertakken I think re-opening https://github.com/webbertakken/unity-builder/issues/77 might be the better approach - as it seems to be a regression

PR #101, opened 2020-06-12
Set android versionCode

benoitdion:
Use action input `androidVersionCode` when provided. Generate the androidVersionCode from the version otherwise.


webbertakken:
Note that this warning is caused by the latest versioning update. This is something I will fix (when I have time) 

The problem is that it should use `this.branch` instead of `origin/${this.branch}` in the versioning script. But this will need to be tested for `push` `pull request` `pull request from downstream` and `tag`.

```
##[warning]describe origin/android-versioncode
fatal: Not a valid object name origin/android-versioncode
```

benoitdion:
@webbertakken I was just looking into that and added some logging. I might need to find a fix or workaround to continue testing.

webbertakken:
You might just remove the `origin/` part right below where you added the log.

benoitdion:
@webbertakken 

(ignore some of the workarounds to get the build going)

the android build is now failing because the gradle daemon is crashing. This usually means there's an OOM. I'm not seeing this on my builds with a very similar configuration. The main difference is we're running `2019.3.13f1`. I see there was an attempt to get `2019.3` builds going but was abandoned due to a license issue. Do you know if it's something we could resume? If so, could you help generate a new license file?

webbertakken:
Hi @benoitdion, 

Can we somehow prove that this is an out of memory issue?

For version 2019.3 builds work fine with the exception of a few specific versions that have the Unity Docker bug (nothing we can do here except picking one of the working versions). In these same issues some versions that should work for sure have been mentioned.

I know of other builds failing because of an out of disk space issue and people have successfully utilised workarounds for this, like cleaning up the runner. As a more scalable option we're also working to delegate the build process to kubernetes, however this is still a work in progress for now. 

I cannot not recall of any attempts where there are out of memory issues or that have been abandoned due to licensing (most people just pick a proven version, also for 2019.3), could you link the issue you're referring to?

Generating a license file should be straight forward, see https://unity-ci.com/docs/github/activation or https://github.com/webbertakken/unity-request-manual-activation-file for this. Or did you mean to have unity builder main repo update to a 2019.3 version? This is something I can do if it is indeed necessary.

benoitdion:
I meant the main repo (see https://github.com/webbertakken/unity-builder/blob/master/.github/workflows/main.yml#L37. We could upgrade all apps, or potentially only android as part of a matrix build. This could also be temporary if we're able to find a better solution to the gradle issues. 

`Gradle build daemon disappeared unexpectedly (it may have been killed or may have crashed)`. In my experience, this usually means a gradle bug or the process ran out of memory. Usually the latter.

webbertakken:
So if I understand correctly, you think the Gradle issue is due to the Unity version?

I'll update the version, but currently I don't understand why you're asking me to do this.

benoitdion:
To be honest, I don't completely understand the issue yet. 

I know the gradle errors we're seeing are usually due to out of memory. 
A similar set up for my own builds works fine with a 2019.3 version of unity. 

It's not clear if upgrading will solve the issue but it feels reasonable to try before digging further.

webbertakken:
Progress can be tracked in https://github.com/webbertakken/unity-builder/pull/102

webbertakken:
@benoitdion it's merged

README.md
```diff
@@ -295,6 +295,12 @@ No version will be set by Builder. **(not recommended)**
 > Not recommended unless you generate a new version in a pre-commit hook. Manually
 > setting versions is error-prone.
 
+#### androidVersionCode
+
+Configure the android `versionCode`.
+
+When not specified, the version code is generated from the version using the `major * 1000000 + minor * 10000 + patch` scheme;
```
webbertakken, line 302: probably we need to make it possible to have more than 99 minor versions, since it's may not be very uncommon in some use cases.

package.json
```diff
@@ -10,12 +10,14 @@
     "prebuild": "yarn",
     "build": "ncc build src --out action --minify",
     "lint": "prettier --check \"src/**/*.js\" && eslint src",
+    "format": "prettier --write \"src/**/*.js\"",
```
webbertakken, line 13: All modern editors have a `run on save` option for prettier nowadays. For the people that forget to enable it, there's a pre-commit hook that'll run for all changed files. 

Knowing that do you think we still need this?

src/model/android-versioning.js
```diff
@@ -0,0 +1,30 @@
...
+
+  static versionToVersionCode(version) {
+    const parsedVersion = semver.parse(version);
+
+    if (!parsedVersion) {
+      return 1;
+    }
+
+    // The greatest value Google Plays allows is 2100000000.
+    // Allow for 4 patch digits, 2 minor digits and 3 major digits.
```
webbertakken, line 19: To me 3-3-4 digits would make more sense than 3-2-4 digits.

src/model/android-versioning.js
```diff
@@ -0,0 +1,30 @@
...
+      return AndroidVersioning.versionToVersionCode(version);
+    }
+    return inputVersionCode;
+  }
+
+  static versionToVersionCode(version) {
+    const parsedVersion = semver.parse(version);
+
+    if (!parsedVersion) {
+      return 1;
```
webbertakken, line 15: should we show a warning here?

package.json
benoitdion: the pre-commit hook runs the check but doesn't fix the issues. I have `format on save` configured but somehow one of the files didn't get formatted properly and I was too lazy to find which one so added this. I can remove if you don't think this is gonna be useful for others or going forward though.

src/model/android-versioning.js
benoitdion: We only have room for 9 digits so unfortunately, that won't work. I prefer 3-2-4 over 3-3-3 because if people don't increment major/minor often, it's very convenient to have 4 digits. 

I use `${{ github.run_number }}` in my builds since it's a bit safer and it's not that important to have `versionCode` correspond to `versionName`. Preferences on this vary though.

src/model/android-versioning.js
benoitdion: Good idea. Initially, I had it fail the build but that's a bit aggressive if someone doesn't build for android. I also thought about only running this logic when the target is android but you'll likely want any of your build to fail early if android version generation isn't gonna work. Returning `1` was a nice balance between the two, adding a warning will make it a bit clearer though.

src/model/android-versioning.js
webbertakken: So if google play allows 2 100 000 000, that looks like 10 digits to me. Could you explain how you derive 9 digits?

My use case would be to have major always at 0 or 1 or 2. Perhaps if the project gets enormously successful i might get up to 20 in several years, but never 99, let alone need more digits than for minor.

> We only have room for 9 digits so unfortunately, that won't work. I prefer 3-2-4 over 3-3-3 because if people don't increment major/minor often, it's very convenient to have 4 digits.

Minor however I use when a new feature gets released. I will almost certainly reach 0.99.0 and 1.99.0. As major digit is reserved for semantically major releases I think 2 digits for minor level is not really safe either.

I would say that depending on the answer to the first question either 2-3-4 or 3-3-4 would be the best compromise. Let me know what you think.

src/model/android-versioning.js
webbertakken: Good thinking. Especially for a first iteration that is perfect. If you add the warning we should be good.

src/model/android-versioning.js
benoitdion: I'd really like to treat the most significant digit as reserved. When you get to 1000000000, that should be a strong warning sign to consider a different versioning scheme to avoid running out of numbers. 

With 3-2-4, I was thinking of apps like Firefox that increment the major monthly. We could also go 3-3-3 and aggressively throw errors when one of the numbers is over 999. In the future, there could also be an option for users to specify how the number is generated.

src/model/android-versioning.js
```diff
@@ -0,0 +1,30 @@
+import * as semver from 'semver';
```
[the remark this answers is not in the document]
benoitdion: Some of the logic in `Versioning` didn't fit exactly what I was trying to do and `semver` is a pretty popular, well-tested library. I can probably remove the dependency if you're against adding this though. There might also be an opportunity to simplify some of the logic in `Versinioning`

src/model/android-versioning.js
benoitdion: I'll start us with `3-3-3`, it's backward compatible with `3-3-4` if you don't give any semantic meaning to the numbers.

>However we're not talking about enterprise software exposing a public API here but about builds for games or graphical programs that are generated by the Unity Editor. Versioning for games to me seems a lot more about having a unique incremental identifier for each build than about anything else.

I'm with you 100% on that. We don't actually use semver for our apps since there's really no consumer of that version and incrementing a number is a simpler mental model. The configuration we're using is

```
versioning: 'Custom'
version: 0.0.${{ github.run_number }}
androidVersionCode: ${{ github.run_number }}
```

src/model/android-versioning.js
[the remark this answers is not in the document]
webbertakken: Yea I think the latter might be worth more in the future. No need to remove it.

benoitdion:
@webbertakken cleaned up git history and addressed review comments. Do you mind taking another look?

webbertakken:
@benoitdion great work and thanks for addressing the comments!

PR #106, opened 2020-06-15
Add support for android builds

benoitdion:
The emoji in the github action name exposed as a env variable through docker was causing a gradle crash 😢.

This PR depends on #105, DO NOT MERGE until after #105 is landed.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,4 +1,4 @@
-name: Actions 😎
+name: Actions
 
 on:
   pull_request: {}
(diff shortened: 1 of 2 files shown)


webbertakken:
This one now has conflicts because of last merge.

.github/workflows/main.yml
```diff
@@ -1,4 +1,4 @@
-name: Actions 😎
+name: Build and Tests
```
webbertakken, line 1: Could you leave this as `actions` just without the emoji? it'll look more readable in this view: 
image

Note that this will also affect the badge in the readme.

benoitdion:
@webbertakken rebased and addressed the feedback. Thanks for taking the time to reviews these!

PR #112, opened 2020-06-25
Document additional action inputs

benoitdion:
This helps avoid warnings when using previously undocumented inputs


webbertakken:
🎉

Issue #34, opened 2020-02-05
Keystore & password


Haelle:
Is there a way to pass the keystore file in the parameters ? And also the password ? (using github secrets)

Should it be in the `customParameters` ?

webbertakken:
@Haelle thanks for your request!

I do not know your current context. Could you tell us a bit more about what you are trying to do? Links to official documentation of related tools would be very helpful too. 

Technically you could pass anything in `customParameters` but it depends on the build method what happens to it.

What would be your suggested approach or desired API?

Haelle:
To deploy to the Google PlayStore you need to [sign your application](https://developer.android.com/studio/publish/app-signing).
It's done with a [keystore](https://developer.android.com/training/articles/keystore.html).

In Unity you can then set your keystore in the [Publishing menu](https://docs.unity3d.com/Manual/class-PlayerSettingsAndroid.html#Publishing).

And everytime you want to build your app you need to provide the password of the keystore via the Editor actually.

So for instance via Unity Cloud build you can provide in their web interface the keystore and the password (cannot show sorry). Here is the info in the [FAQ](https://developer.cloud.unity3d.com/support/faq/) section Android, 2rd question

I've never worked with the command line tool, so currently I have no idea how you're doing the command line build... I suppose there is a way to sign the app (as Unity Cloud Build does it !)

darkgap:
Looking at [here](https://github.com/GabLeRoux/unity3d-ci-example#how-to-add-build-targets), the [GabLeRoux](https://github.com/GabLeRoux/unity3d-ci-example#how-to-add-build-targets) unity3d library lets you specify the keystore by binary64  formatting it and putting it in a repos' secret variable

dabo248:
@Haelle did you find a solution for this?

webbertakken:
Feel free to open a PR with your suggested approach. You're free to make any additions to the builder method, as long as it's usable for everyone.

Haelle:
This is in pending on my side, I'll come back to it later

webbertakken:
@Haelle did you work this out? Anything we can reuse for everyone?

Haelle:
No sorry CI is postponed indefinitely for the moment

davidmfinol:
Added these parameters: `androidAppBundle`, `androidKeystoreName`, `androidKeystoreBase64`, `androidKeystorePass`, `androidKeyaliasName`, `androidKeyaliasPass`

[...]

PR #115, opened 2020-07-06
Add Android Build Settings and macOS buildNumber

davidmfinol:
Adds parameters for Android for #34:
-androidAppBundle 
-androidKeystoreName
-androidKeystoreBase64
-androidKeystorePass
-androidKeyalias
-androidKeyaliasPass

Need to update PlayerSettings.macOS.buildNumber for each submission to the Mac App Store. Per [Unity ](https://docs.unity3d.com/Manual/HOWTO-PortToAppleMacStore.html) and [Apple ](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102364) documentation, just applying the auto-generated version makes the most sense.


davidmfinol:
@webbertakken 
I think the Tests check failed only because of some transient problem? At the least, I don't see a problem with the tests themselves.

I do have another problem with this PR, though. I'm currently getting an error when submitting to the Mac App Store: 
[Transporter Error Output]: ERROR ITMS-90061: "This bundle is invalid. The value for key CFBundleVersion [1.30.6] in the Info.plist file must contain a higher version than that of the previously uploaded version [38]. Please find more information about CFBundleVersion at https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleversion"

My build number is currently at 38 since I had been manually incrementing the number by 1, similar to what I had been doing for the Android releases (a mistake, in hindsight). 
I can't use the the androidVersionCode: Apple will interpret that number as the first number for CFBundleVersion, which is limited to 4 digits, while the androidVersionCode takes 7 digits with how we currently have it. 

This PR may not be the appropriate place to ask, but I figured I'd ask for any suggestion on what I should do?

webbertakken:
See contribution [guide](https://github.com/webbertakken/unity-builder/blob/master/CONTRIBUTING.md).

Specifically this part:
> Please note that commit hooks will run automatically to perform some tasks;
>
> - format your code
> - run tests
> - build distributable files

The test fails because the distributable file is not automatically generated. Did the pre-commit hook fail?

davidmfinol:
Pre-commit hook ran fine for me.

I just re-ran yarn build again, and I'm not seeing an error: 
image

webbertakken:
Right. It was actually yarn getting a 502. I reran the tests and now they are green.

mastef:
@davidmfinol could `androidCreateSymbolsZip` be implemented in the same way?

davidmfinol:
Yeah, I think you should be able to raise a PR for it.
I think you would need to add `androidCreateSymbolsZip` to `action.yml`, `input.ts`, `build-parameters.ts`, `docker.ts`, `build.sh`, and `AndroidSettings.cs`.
We'd also want to add it to [the documentation](https://github.com/game-ci/documentation/blob/main/docs/github/v2/04-builder.md).


I actually wonder if there is a better approach that would easily enable us to set all of the [EditorUserBuildSettings](https://docs.unity3d.com/ScriptReference/EditorUserBuildSettings.html) through our default build script. That's probably a question for another time, though.

mastef:
@davidmfinol how about this : 

```diff
      - uses: game-ci/unity-builder@v2.0-alpha-6
        name: Build unity project to Android
        with:
           unityVersion: ${{ matrix.unityVersion }}
           targetPlatform: ${{ matrix.targetPlatform }}
           buildsPath: build-${{ matrix.targetPlatform }}
           buildName: ${{ secrets.BUILD_NAME }}
           versioning: Custom
           version: ${{ steps.versionstep.outputs.buildVersion }}
           androidVersionCode: ${{ steps.versionstep.outputs.buildNumber }}
           androidAppBundle: true
           androidKeystoreName: ${{ secrets.ANDROID_KEYSTORE_NAME }}
           androidKeystoreBase64: ${{ secrets.ANDROID_KEYSTORE_BASE64 }}
           androidKeystorePass: ${{ secrets.ANDROID_KEYSTORE_PASS }}
           androidKeyaliasName: ${{ secrets.ANDROID_KEYALIAS_NAME }}
           androidKeyaliasPass: ${{ secrets.ANDROID_KEYALIAS_PASS }}
+          customParameters: -androidCreateSymbolsZip
```

davidmfinol:
You can do it that way if you want to quicker solution for your project. You would also need a custom `buildMethod` that applies `EditorUserBuildSettings.androidCreateSymbolsZip = true;`

mastef:
> You can do it that way if you want to quicker solution for your project. You would also need a custom `buildMethod` that applies `EditorUserBuildSettings.androidCreateSymbolsZip = true;`

Somebody else also tried with that setting and no success https://answers.unity.com/questions/1767637/batchmode-command-not-generating-symbolszip-file.html?childToView=1822466#answer-1822466 but maybe they didn't add the setting in the cli.

So @davidmfinol you're saying both settings are required? The `EditorUserBuildSettings.androidCreateSymbolsZip = true` and additionally the `-androidCreateSymbolsZip` flag in the CLI? That could be a solution to his problem then as well

davidmfinol:
Passing `-androidCreateSymbolsZip` in the CLI could be useful to check if that flag was passed in your build script, but it's not necessary.

My understanding is that only `EditorUserBuildSettings.androidCreateSymbolsZip = true` is required, so I'm not sure why it didn't work for him.

mastef:
> Passing `-androidCreateSymbolsZip` in the CLI could be useful to check if that flag was passed in your build script, but it's not necessary.
> 
> My understanding is that only `EditorUserBuildSettings.androidCreateSymbolsZip = true` is required, so I'm not sure why it didn't work for him.

Because for me - passing it in the CLI - is also not sufficient. There may be a bigger issue at play here :|

PR #119, opened 2020-07-08
Log git diff during build

### LINKED ISSUE #117: Show diff for dirty repository?
Hi, I have been trying out the actions and got a workflow functional with allowDirtyBuild: true. Really cool stuff, thank you for making these actions. I had to add /[Aa]rtifacts/ to .gitignore and fix an issue that was my fault based on inconsistent filename case to reduce the amount of dirty changes in the repo. But it's still showing `ProjectSettings/ProjectSettings.asset` as changed. 

```
 M ProjectSettings/ProjectSettings.asset

##[error]Branch is dirty. Refusing to base semantic version on uncommitted changes
```

I'm happy to get any help for determining why ProjectSettings is changing, but maybe in general it would be nice to have an option like `showDiffOnDirty: true` to show what changes have happened since checkout.

I know there could potentially be many files changed in a dirty repository scenario, but just knowing the filenames isn't quite enough information for me to diagnose the issue.

Related: #95 

Edit: This can be achieved with nice output by issuing `git status --verbose --verbose` or `git --no-pager diff` (conditionally if this option is enabled)

dogboydog:
* Issue `git --no-pager diff` if logDiffifDirty is specified for a build, whether or not you allow dirty builds. 
* I know "DiffIf" looks a little hard to read. I can refactor it to a different name if you have a better idea.
* allowDirtyBuild was missing from action.yml. I don't know if this was intentional, but I added it with the information from README.md
* basic mocking tests added

Working screenshot: 

image

Side note... any idea why  ProjectSettings.asset is being modified during the build? I am using the same Unity version in the branch as I am specifying to the Unity Actions.  It seems like the defaults used by unity-builder override the ProjectSettings. I wonder if you should ignore "dirtiness" if the only dirty file is ProjectSettings if that's the case.

Fix #117


dogboydog:
Thanks. 

> I think we should come to some common sense amount of files or lines of which we'd show the diff and that way prevent having to add another parameter to the workflow.
>
>My suggestion would be to always show the first 20~100 lines of diff across files.

Okay, sure. So when you say across files, do you mean 20-100 lines total? Not that mine lines per file? Would this number be configurable (like via an environmental variable) or just a constant?

> This would be prevented too by the suggestion, but otherwise i'd probably have suggested something like logDifference or logDiff (as the idea was not always log it either way).

Now that you mention it, the diff would just be blank if the repository were not dirty, so just logDiff would still seem to explain the concept to me and be much easier to read.

webbertakken:
> Okay, sure. So when you say across files, do you mean 20-100 lines total? Not that many lines per file? Would this number be configurable (like via an environmental variable) or just a constant?

Yes I meant in total, because we have to weight log pollution vs ease of use.

I wouldn't make this number configurable, at least not yet. For your use case this solves the problem and I think it will for most other cases where only 1 or 2 files change too. Making things variable would only make it harder to maintain, so I'd suggest we wait for a valid use case for it.

dogboydog:
Cool. My plan is to refactor the option to be "logDiff", and limit it to whatever number of lines you want. 60?

webbertakken:
> Cool. My plan is to refactor the option to be "logDiff", and limit it to whatever number of lines you want. 60?

I still think we shouldn't expose an extra parameter but just enable this functionality by default.

dogboydog:
Oh I see. I misread your first message, sorry.  Okay I will do it unconditionally and remove the parameter.

dogboydog:
Okay, I have your suggestions working. Diff lines are capped at 60 for now. Let me know what you think. Thanks for the feedback.

Beginning of diff:

image
End of diff:

image


I can move these questions to a different issue if you want, but I was curious about a few things.

* I have working code (pretty simple) to read `m_EditorVersion` from `ProjectSettings\ProjectVersion.txt`. I'm using this as the input to `with: unityVersion`. Something like this could be the default if unityVersion is not specified. Any thoughts on that? Seems to be in keeping with the goal of having things work out of the box, but maybe you want to avoid parsing this file? Maybe in production people want to use a constant version to build binaries, but you'd think they'd be developing with that version too, right? It looks like the ProjectVersion.txt file has been around for a while, and is automatically updated by Unity when you upgrade. Just one idea

* Should there be certain files excluded from consideration for "dirtiness"?  I'm not sure how I can avoid ProjectSettings.asset being modified during the build since I am just using the default settings for the action. Although depending on the buildMethod specified, maybe other files like .meta files could change. I can just keep using allowDirtyBuild for the moment either way though.

webbertakken:
> I can move these questions to a different issue if you want

Yes I think that would be better

webbertakken:
https://github.com/webbertakken/unity-builder/issues/95

PR #125, opened 2020-07-18
Pin to working unity-3d image

benoitdion:
After initial investigation, it seems https://gitlab.com/gableroux/unity3d/-/commit/4d1eaa74cbec79386cda4f797e1db63ebcda1c20 and related commits broke the unity-builder android builds. Unity is no longer able to find the Android Tools or the JDK.

I was able to modify the Dockerfile to export an `ANDROID_HOME` env variable to get past the first issue but so far haven't found how to configure the Unity JDK location. Setting a `JAVA_HOME` env variable doesn't seem to help.

Until we find a fix upstream (https://gitlab.com/gableroux/unity3d), pinning to an older version of the unity-3d image. We probably don't want to merge this since the image hash isn't going to work on all platforms.

Corresponding issue at gitlab unity-3d: https://gitlab.com/gableroux/unity3d/-/issues/82


webbertakken:
Related: https://gitlab.com/gableroux/unity3d/-/issues/85

We're looking for good ways to make Android images run well in the future across all versions.

webbertakken:
Android has been reported to work since 2.0-alpha versions. Closing this.

PR #129, opened 2020-07-24
Fix boolean logic

benoitdion:
Fix an issue where `androidAppBundle` was always true because the string `'false'` is truthy.


action.yml
```diff
@@ -40,7 +40,7 @@ inputs:
     description: 'The android versionCode'
   androidAppBundle:
     required: false
-    default: false
+    default: 'false'
```
webbertakken, line 43: input is always parsed as a string as far as i know

webbertakken:
As everything in the Input model assumes string as both input and output for all parameters, I don't think it's a good idea to change this just for this parameter, if at all.

benoitdion:
>However what we could do is leave the input parameters to the Input model as strings (as per the actions core/input spec) and return values from the Input model to what makes sense for each parameter.

It looks like what we were trying to do for `androidAppBundle` and `allowDirtyBuild` but the logic was incorrect and led to the issue I'm trying to fix with this PR. I do think it's important to convert the strings to booleans as early as possible otherwise we're gonna continue to run into bugs like this, especially when using untyped javascript.

webbertakken:
> > However what we could do is leave the input parameters to the Input model as strings (as per the actions core/input spec) and return values from the Input model to what makes sense for each parameter.
> 
> It looks like what we were trying to do for `androidAppBundle` and `allowDirtyBuild` but the logic was incorrect and led to the issue I'm trying to fix with this PR. I do think it's important to convert the strings to booleans as early as possible otherwise we're gonna continue to run into bugs like this, especially when using untyped javascript.

Fair point. We'll merge as soon as android builds are stable again.

webbertakken:
FYI, we have disabled the Android builds in the example workflow while https://gitlab.com/gableroux/unity3d/-/issues/85 is still in progress. Feel free to rebase on master and keep merging things.

benoitdion:
@webbertakken rebased!

webbertakken:
Those two failures are expected, since the secrets are not exposed to external forks. We'll merge without them passing.

PR #133, opened 2020-08-09
Cleanup root dir

webbertakken:
- remove package-lock.json
- move workflow scripts into action (to be incorporated with it later)


benoitdion:
why is the lock file being removed?

webbertakken:
Because we've been using yarn since the initial commit. Yarn uses `yarn.lock` instead.

benoitdion:
oh, duh!

PR #137, opened 2020-08-15
Rename custom -version flag to -buildVersion

### LINKED ISSUE #136: 2019.4.8f1 builds are empty, but reported as success
After upgrading my Unity version (with no other changes to my use of `unity-actions`), now build output looks like this:
```
###########################
#    Building platform    #
###########################

2019.4.8f1 (60781d942082)
Build succeeded

###########################
#     Build directory     #
###########################

total 8.0K
drwxr-xr-x 2 root root 4.0K Aug 15 13:25 .
drwxr-xr-x 3 root root 4.0K Aug 15 13:25 ..
```

I encountered this by using `unityVersion: 2019.4.8f1`, but you can reproduce the underlying issue more directly by trying the `-version` flag at the command line:

```sh
$ docker pull gableroux/unity3d:2019.4.8f1-linux-il2cpp
$ docker run -it --rm gableroux/unity3d:2019.4.8f1-linux-il2cpp
# /opt/Unity/Editor/Unity -batchmode -version "v1.0" -logfile /dev/stdout
2019.4.8f1 (60781d942082)
```

It appears that `-version` is interpreted by the Unity CLI to print its own version, rather than provide a version for the build.

When it's not provided, Unity continues to start up as normal:

```sh
# /opt/Unity/Editor/Unity -batchmode -logfile /dev/stdout
[LicensingClient] ERROR Failed to connect to local IPC
[Licensing::Module] Failed to connect to channel: LicenseClient-root
[Licensing::Module] Successfully launched the LicensingClient
[SignatureVerifier] Application signature verification not supported on this platform.
[Licensing::Module] Successfully connected to LicensingClient on channel: LicenseClient-root
[Licensing::Module] Successfully launched and connected to LicensingClient
Entitlement-based licensing initiated
[LicensingClient] Licenses Updated successfully in LicensingClient
Error opening default X display
(Filename: ./Editor/Platform/Linux/X11Quarantine.cpp Line: 33)
```

(This failure is expected, as I didn't set up an X server for this test.)

I couldn't find any mention of a `-version` flag [even in the 2019.3 docs](https://docs.unity3d.com/2019.3/Documentation/Manual/CommandLineArguments.html), so I'm not sure what the best solution is here.

Many thanks for this project! It's been exceptionally handy for a game side project. 😄

jspahrsummers:
`-version` is now used (as of 2019.4.8f1 or earlier) by the Unity CLI to print its own version, then exit with code 0. This interferes with `unity-builder`'s version passing, so I propose renaming the argument to something less likely to collide.

I was unable to run Yarn as requested in the contributing guidelines, as I'm missing tons of prerequisites. Besides whatever CI checks this repo can provide, I can confirm that this fixed #136 in my game's workflow.

### DIFF
--- a/action/default-build-script/Assets/Editor/Builder.cs
+++ b/action/default-build-script/Assets/Editor/Builder.cs
@@ -27,7 +27,7 @@ namespace UnityBuilderAction
       };
 
       // Set version for this build
-      VersionApplicator.SetVersion(options["customProjectVersion"]);
+      VersionApplicator.SetVersion(options["buildVersion"]);
       VersionApplicator.SetAndroidVersionCode(options["androidVersionCode"]);
       
       // Apply Android settings
--- a/action/default-build-script/Assets/Editor/Builder.cs
+++ b/action/default-build-script/Assets/Editor/Builder.cs
@@ -27,7 +27,7 @@ namespace UnityBuilderAction
       };
 
       // Set version for this build
-      VersionApplicator.SetVersion(options["version"]);
+      VersionApplicator.SetVersion(options["customProjectVersion"]);
       VersionApplicator.SetAndroidVersionCode(options["androidVersionCode"]);
       
       // Apply Android settings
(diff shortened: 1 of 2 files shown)


webbertakken:
This looks good to me.

Though I think it's better to call it `buildVersion` instead, 
without custom as it's not necessarily overriding the actions default version, which is also called customVersion. 
and `build` instead of `project` as that is a more correct scope.

If you change that we'll merge it. (Don't worry about the k8s checks failing, they're not required)

jspahrsummers:
Sure, makes sense, thanks! Now renamed to `-buildVersion`.

PR #139, opened 2020-08-22
Use provision and dispose GKE action - K8s cleanup

frostebite:


webbertakken:
@frostebite are you still working on this? should it be merged?

PR #141, opened 2020-08-22
Fix file ownership issues for runners.

xanantis:
Fixes issues with file ownership after docker usage. 
For Github Runners,  it is required to call "sudo chown" to fix file ownership.
For Self-hosted Runners, it creates problems, if the user is not allowed to call sudo. One of the hacky solutions is to wrap the docker executable.

So, it is better, to build and run docker as the current user.


webbertakken:
This was reverted in https://github.com/game-ci/unity-builder/pull/165 and two new solutions came up:

- Chowning after build run: https://github.com/game-ci/unity-builder/pull/250
- Setting docker image to non-root user https://github.com/game-ci/docker/pull/106

Follow developments there, or in their followups.

PR #164, opened 2020-10-20
refactored to use unity-editor alias

BLaZeKiLL:
#### Changes

- refactored to `unity-editor` alias to conform with the new docker images
- hard coded repo version as `0` in both `image-tag.js` & `image-tag.jest.js`

#### Checklist


GabLeRoux:
Considering this will require new activation, I think we can consider this as a breaking change and we could bump the major version. @webbertakken what do you think?

webbertakken:
Agreed. It'll also probably have a few bugs at first, because a lot has changed in the images. We might even go with a `2.0.0-prerelease-1` or something

action/steps/build.sh
```diff
@@ -108,24 +108,24 @@ echo "#    Building platform    #"
...
-    -buildVersion "$VERSION" \
-    -androidVersionCode "$ANDROID_VERSION_CODE" \
-    -androidKeystoreName "$ANDROID_KEYSTORE_NAME" \
-    -androidKeystorePass "$ANDROID_KEYSTORE_PASS" \
-    -androidKeyaliasName "$ANDROID_KEYALIAS_NAME" \
-    -androidKeyaliasPass "$ANDROID_KEYALIAS_PASS" \
-    $CUSTOM_PARAMETERS
+unity-editor \
+  -batchmode \
+  -nographics \
```
webbertakken, line 113: Would you be ok to put `-nographics` in a separate PR? otherwise this is not manageable from a maintenance perspective.

webbertakken:
This will require a few more changes. 

In the javascript part the images will have to come from `unityci/editor` instead of `gableroux/unity3d`. And it might involve a little bit of inclusive coding there, as I don't know what else will break.

action/steps/build.sh
BLaZeKiLL: okay, will update it

BLaZeKiLL:
> This will require a few more changes.
> 
> In the javascript part the images will have to come from `unityci/editor` instead of `gableroux/unity3d`. And it might involve a little bit of inclusive coding there, as I don't know what else will break.

I was testing with the custom image flag so didn't think of that, will take a look at the javascript code and see what all will change

PR #165, opened 2020-10-22
Fix android build workaround

caiusno1:
#### Changes

- Revert File ownership PR (Fix file ownership issues for runners. #141)

Reasons:
To my best knowledge enabling the docker container to run with runner rights instead of sudo rights require considerations/iterations on the build of the base images from https://gitlab.com/gableroux/unity3d. Android sdkmanager seems to have files created in the user directory of the user that was active on the build of the base image (which was sudo by default). Changing the user in the build of the dependent docker image (unity-builder) results in Unity to search for these files in the user directory of the current user (where they are not available). I am not deep enough in android development to know whether one can simply copy these files from root diretory to user directory or delegate the sdkmager to search in root instead of user directory but both feels not to be a good practice also. I would suggest that we maybe branch from current state so that PR File ownership PR "(Fix file ownership issues for runners " can be readded when capability to support runner user instead of root is given from the base images. Btw. the "gradle daemon disappeared unexpectedly" error seems to be created by non ascii chars in the workflow file of my unity project. Maybe the Get Started example sample should be changed to address that issue :D @webbertakken .

I hope you understand and share my opinion on this situation. In case of a question to these considerations please write me a message :).

#### Checklist


webbertakken:
Might now be fixed in https://github.com/game-ci/unity-builder/pull/250

PR #206, opened 2021-01-26
Remove debug from versioning

davidmfinol:
Would like to remove this debug option, as I don't like how it generates many unnecessary warnings, and I don't think it provides any value.

#### Changes

- Remove these:
image

#### Checklist


GabLeRoux (approved):

webbertakken:
In case that it fails it may still be useful to give that output. This is why it was added. Maybe only show it on failure?

If there's no nice way to do that feel free to merge it indeed.

PR #235, opened 2021-03-28
AWS 

frostebite:
This PR implements support for AWS Fargate with Git LFS support.

# What it does?
With this feature you can run builds on the AWS Cloud on demand. Most importantly you can run builds for very large projects.

# Why we need it
GitHub Actions run on build machines provided by GitHub. You can see info about the specifications of the build machines here:
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

For game development and Unity projects the disc size in particular is quite small. You may well experience an error related to running out of disc space. You may also want to run the build on a server with more memory or CPU cores.

You can use this feature to send a build from any supported git platform to any of the supported remote build cluster types, now including AWS. This is an inexpensive and reliable way to scale to larger builds. There is virtually no limit to the disc size for the supported remote build cluster types.

# To Do
- docs (in progress)
- steam deploy (estimating start Mon: 12th April 2021)

# Planned for later
- Align parameter features and code (local, k8s and aws all work a bit differently now)
- AWS builds for Git using direct webhook (build any git based unity project)
- Offer cache token parameter rather than assuming by branch/tag
- Automated way to cache all docker images to AWS container registry (ECR)
- Improved test coverage
- Asynchronous build support
- Unity test framework support
- Make fargate task running more reusable as per our needs as a library or tool

# Concept
You can run Game CI Unity builds on the supported platforms. You can also run the build on a remote build cluster such as Kubernetes or a Cloud Provider like AWS. This is a bit like self-hosting build servers except on demand.

The main reasons for doing this are reducing build machine costs, running builds too large for other options, nearly unlimited parallel builds and workflow improvements from further isolating the builds.

There can also be significant performance and cost improvements if you run several builds per commit (or any other event like a PR) because you never have to worry about blocking other builds with this approach.

**Advantages**
- Configure CPU, memory/RAM and disc space as needed
- Pay for usage
- Scale up to as many builds as you want
- No servers to manage or worry about

**Disadvantage**
- cheap but costs aws fees
- requires an AWS account
- additional startup time to begin the build

# Parameters
StackName : The base stack setup in the getting started section.
Memory : 
CPU : 
As disc size is completely elastic there is no need to specify disc size :

## Build artifacts
Completed builds are zipped then uploaded to aws S3. Steam deployment is in progress.

## Logs
Logs are streamed via Kinesis to allow fast and stable log streaming, with no aws polling issues.

## Cleaning up resources
Even in the event of an unexpected error we have a failsafe to make sure resources don't accidently remain active and charging you. This is accomplished with a lambda function that runs after a delay if the resources are still active.


src/index.ts
```diff
@@ -10,14 +10,27 @@ async function run() {
...
+      case 'aws':
+        core.info('Building with AWS');
+        await AWS.runBuildJob(buildParameters, baseImage);
+        break;
+
+      // default and local case
+      default:
+        core.info('Building locally');
+        // Build docker image
+        // TODO: No image required (instead use a version published to dockerhub for the action, supply credentials for github cloning)
```
webbertakken, line 30: Do these comments belong to case 'aws'? Or how should we read this?

frostebite: Cleaned up, was an old message now

jsteinich:
I'm curious if you've run into timeout issues launching fargate tasks. It's been a couple years since I've tried, but I had ran into issues due to the size of the container image (especially for Android).

frostebite:
Nope, it's been working fantastically.

PR #238, opened 2021-03-31
Fix build script for Unity 2018, by using reflection instead of version defines

### LINKED ISSUE #233: Asmdef with conditional errors out on editor <= 2018
**Bug description**

We created an assembly definition to support addressables in https://github.com/game-ci/unity-builder/pull/231

We didn't account for 2018 erroring out when the builder assembly definition references the addressables assembly in case it is not present (which works fine for 2019+ versions).

**How to reproduce**

- Run a build on 2018 without addressables in the project

**Expected behavior**

I will quote @sbergen here:

> > > All 2018.3.14f1 tests are failing with `Assembly has reference to non-existent assembly 'Unity.Addressables.Editor' (Assets/Editor/Editor/UnityBuilderAction/UnityBuilderAction.asmdef)`. I don't believe that is related to this PR. Instead, I'm thinking it is caused by [game-ci/unity-builder#231](https://github.com/game-ci/unity-builder/pull/231)
> > 
> > 
> > Yes. This is not related to your PR indeed.
> > @sbergen any ideas on how to fix this?
> 
> Hmm, I'm thinking that they only made that not-an-error when they introduced version defines in 2019. Didn't think of this 😞
> 
> Unfortunately, you can't have anything like `#if UNITY_2019...` in an .asmdef, so I can't think of a really elegant solution.
> 
> The options I can think of are:
> 
> * Making two versions of the .asmdef: one for >= 2019, and another for older versions
> * Reverting back to not using an asmdef, and making that logic be behind a builder option, instead of using version defines. You'd then define some symbol from the build script for whether or not the logic should be included.
> * Parsing `manifest.json` manually, and using that to define a symbol, instead of doing it based on an option.

**Additional details**

Personally I like the first two options. 

In case of the first option we could just make one `.asmdef` if we drop support for 2018 right?

Which of the solutions would you recommend?

sbergen:
#### Changes

- Use reflection instead of version defines to see if the addressables package is installed.

Fixes #233

#### Checklist

### DIFF
--- a/dist/default-build-script/Assets/Editor/UnityBuilderAction/Builder.cs
+++ b/dist/default-build-script/Assets/Editor/UnityBuilderAction/Builder.cs
@@ -1,15 +1,12 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using UnityBuilderAction.Input;
 using UnityBuilderAction.Reporting;
 using UnityBuilderAction.Versioning;
 using UnityEditor;
 using UnityEditor.Build.Reporting;
-
-#if USE_ADDRESSABLES
-using UnityEditor.AddressableAssets.Settings;
-#endif
+using UnityEngine;
 
 namespace UnityBuilderAction
 {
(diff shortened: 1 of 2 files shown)


davidmfinol (approved):

webbertakken:
Great work! Thank you very much for fixing it across the board.

PR #247, opened 2021-04-14
Fixes files owned by root

### LINKED ISSUE #219: Files created by the build are owned by root, not "runner"
**Bug description**

After building using the unity-builder github actions, the output files are owned by `root.root` instead of `runner.docker` like the files in git.

**How to reproduce**

Build.

**Expected behavior**

Files can be modified by the github actions

**Additional details**

The root cause here is that my build doesn't build to build/${{ targetPlatform }}/${{ targetPlatform}}.$EXTENSION, but somewhere else, and so I want to move the output around afterwards in the github workflow rather than in the build function, but I can't because of permissions. But it seems like, in general, this would be a better practice to run `chown` within the docker before finishing.

Bradshaw:
Files created by the build are owned by root, since the internal docker is running as root

Should close #219 

This can be fixed by providing the Dockerfile with a UID and GID to use instead
Doing this causes an issue where the Android SDK is inaccessible, so it's necessary to also chown the /opt/unity folder to the runner user

#### Changes

- Creates a user named `runner` with UID and GID 1000 in `dist/Dockerfile`
- Sets `runner` as the internal docker container user
- `chown`s and `chmod`s the necessary files and folders for `runner` to be able to build

#### Checklist

- [x] Readme not needed
- [x] Tests not needed (I think)
- [ ] Fix the `no space left on device` error that pops up in [the pull-request check build](https://github.com/game-ci/unity-builder/pull/247/checks?check_run_id=2344541140#step:4:220)

#### Explanation

Most of the changes are based on the now-reverted changes in #141:

- In `dist/Dockerfile`, a new user, `runner` is created with UID and GID 1000
- A `/github` directory is created and `chown`ed to `runner`
- `runner` is set as the user inside the docker contained
- The UID and GID are also added to `src/index.ts`, `src/model/build-parameters.ts` and `src/model/docker.ts`

However, for some reason, just doing this breaks the `activate.sh` step, as for some reason, despite `useradd`ing `runner` with `/github/home` as their home directory, the following scripts see an empty `$HOME` variable.   
This is fixed by adding `export HOME=/github/home` to the start of `activate.sh`

Running the build like this fixes the original issue. The build succeeds and all the files are owned by `runner` instead of root, however, this breaks Android builds (c.f: #165)  

One of the errors that pop up is the following:
```
Win32Exception: ApplicationName='/opt/unity/Editor/Data/PlaybackEngines/AndroidPlayer/SDK/tools/bin/sdkmanager', CommandLine='--list', CurrentDirectory='/github/workspace', Native error= mono-io-layer-error (5)
```

Searching that error leads to an explanation that the SDK can't be used because the user doesn't have the correct permissions in the Unity editor folder.  

Adding a `RUN chown $UID:$GID -R /opt/unity/` statement to `dist/Dockerfile` solves this issue.

### DIFF
--- a/dist/Dockerfile
+++ b/dist/Dockerfile
@@ -1,6 +1,10 @@
 ARG IMAGE
 FROM $IMAGE
 
+ARG UNAME=runner
+ARG UID=1000
+ARG GID=1000
+
 LABEL "com.github.actions.name"="Unity - Builder"
 LABEL "com.github.actions.description"="Build Unity projects for different platforms."
 LABEL "com.github.actions.icon"="box"
@@ -10,6 +14,12 @@ LABEL "repository"="http://github.com/webbertakken/unity-actions"
 LABEL "homepage"="http://github.com/webbertakken/unity-actions"
 LABEL "maintainer"="Webber Takken <webber@takken.io>"
 
+RUN bash -c 'mkdir -p /github/{home,workflow,workspace}' && chown $UID:$GID -R /github/
+RUN getent group $GID || groupadd -g $GID $UNAME
+RUN id -u $UID &>/dev/null || useradd -m -u $UID -g $GID -s /bin/bash -d /github/home $UNAME
+
+RUN chown $UID:$GID -R /opt/unity/
+
 ADD default-build-script /UnityBuilderAction
 ADD steps /steps
 RUN chmod -R +x /steps
@@ -18,3 +28,5 @@ RUN chmod +x /entrypoint.sh
 RUN ls
 
 ENTRYPOINT ["/entrypoint.sh"]
+
+USER $UID:$GID
--- a/dist/steps/activate.sh
+++ b/dist/steps/activate.sh
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+echo "Exporting HOME as /github/home"
+export HOME=/github/home
+
 if [[ -n "$UNITY_LICENSE" ]] || [[ -n "$UNITY_LICENSE_FILE" ]]; then
   #
   # PERSONAL LICENSE MODE
--- a/src/index.ts
+++ b/src/index.ts
@@ -26,7 +26,13 @@ async function run() {
       // default and local case
       default:
         core.info('Building locally');
-        builtImage = await Docker.build({ path: actionFolder, dockerfile, baseImage });
+        builtImage = await Docker.build({
+          path: actionFolder,
+          dockerfile,
+          baseImage,
+          uid: buildParameters.uid,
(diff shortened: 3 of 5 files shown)


Bradshaw:
Hmm, apparently the `RUN chown $UID:$GID -R /opt/unity/` step causes an issue...

Maybe don't merge this yet?

https://github.com/game-ci/unity-builder/pull/247/checks?check_run_id=2344541140#step:4:220
```
Step 16/24 : RUN chown $UID:$GID -R /opt/unity/
 ---> Running in 44b0a134ec7c
Error processing tar file(exit status 1): write /opt/unity/Editor/Data/PlaybackEngines/AndroidPlayer/NDK/toolchains/x86_64-4.9/prebuilt/linux-x86_64/lib/lib64/libc++.so: no space left on device
```

Bradshaw:
Why would it cause a `no space left on device` error though?

webbertakken:
These are great steps in the right direction. Thanks for documenting each step on the way. It's very valuable.

I'm thinking perhaps we might need changes in the docker image itself as well.

Amazing work! Looking forward to merge this when it's stable.

Bradshaw:
Marking this as a draft, since it actually breaks build processes and saturates hard drives

I think I'm on the right path, but I need to figure out why `chown`ing `/opt/unity` causes the hard drive to fill up like that

Does `chown` follow symlinks? Maybe there's a loop? 🤷

Bradshaw:
The tests will probably fail for Android, but ought to work if [game-ci/docker pull #106](https://github.com/game-ci/docker/pull/106), which also enables a runner user upstream, is merged

I have built some docker base, hub and editor images [available on my dockerhub](https://hub.docker.com/u/gaeelbradshaw) and have been testing them like so:
```yaml
- uses: Bradshaw/unity-builder@fix/#219/files-owned-by-root
  env:
    UNITY_SERIAL: ${{ secrets.UNITY_SERIAL }}
  with:
    customImage: gaeelbradshaw/editor:ubuntu-${{ matrix.unityVersion }}-android-0.12.0
```

Note, I have only built base and android images for the following Unity editor versions:
- 2019.2.11f1
- 2019.3.15f1
- 2019.4.8f1
- 2019.4.23f1
- 2020.1.7f1
- 2020.2.1f1
- 2020.3.2f1
- 2021.1.4f1

Let me know if you want me to build more versions to test against

Bradshaw:
Closing this because it is possible (and desirable) to fix this all upstream (*c.f:* [game-ci/docker pull #106](https://github.com/game-ci/docker/pull/106))

PR #250, opened 2021-05-01
Chown files to a custom user after building (fix file ownership)

### LINKED ISSUE #219: Files created by the build are owned by root, not "runner"

Bradshaw:
#### Changes

Fixes #219, without causing the issues caused by #141 

- Adds a `chownFilesTo` input
- `chown`s the files to whatever the input string is

Use it like this:
```Dockerfile
- uses: Bradshaw/unity-builder@main
  env:
      UNITY_SERIAL: ${{ secrets.UNITY_SERIAL }}
  with:
      chownFilesTo: "1000:1000"
```

And it does this after building:
```sh
if [[ -n "$CHOWN_FILES_TO" ]]; then
  if [[ -n "$BUILD_PATH_FULL" ]]; then
    chown -R $CHOWN_FILES_TO $BUILD_PATH_FULL
  fi
  if [[ -n "$UNITY_PROJECT_PATH" ]]; then
    chown -R $CHOWN_FILES_TO $UNITY_PROJECT_PATH
  fi
  if [[ -n "$CUSTOM_BUILD_PATH" ]]; then
    chown -R $CHOWN_FILES_TO $CUSTOM_BUILD_PATH
  fi
fi
```

That's it... That's the fix 😅

I still think the internal Docker user should be something other than root, but that requires a heavier refactor and breaks a whole bunch of backwards-compatibility.

This is entirely optional and does absolutely nothing if the `chownFilesTo` input is unset or left blank

#### Checklist

- [x] Readme not needed
- [x] Tests not needed
- [ ] Documentation needs updating, probably

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -109,7 +109,7 @@ inputs:
   chownFilesTo:
     required: false
     default: ''
-    description: After building, change the ownership of the created files to this user
+    description: 'User and optionally group (user or user:group or uid:gid) to give ownership of the resulting build artifacts'
   allowDirtyBuild:
     required: false
     default: ''
--- a/action.yml
+++ b/action.yml
@@ -106,6 +106,10 @@ inputs:
 
       Parameters must start with a hyphen (-) and may be followed by a value (without hyphen).
       Parameters without a value will be considered booleans (with a value of true).
+  chownFilesTo:
+    required: false
+    default: ''
+    description: After building, change the ownership of the created files to this user
   allowDirtyBuild:
     required: false
     default: ''
--- a/dist/steps/build.sh
+++ b/dist/steps/build.sh
@@ -145,15 +145,8 @@ fi
 
 
 if [[ -n "$CHOWN_FILES_TO" ]]; then
-  if [[ -n "$BUILD_PATH_FULL" ]]; then
-    chown -R $CHOWN_FILES_TO $BUILD_PATH_FULL
-  fi
-  if [[ -n "$UNITY_PROJECT_PATH" ]]; then
-    chown -R $CHOWN_FILES_TO $UNITY_PROJECT_PATH
-  fi
-  if [[ -n "$CUSTOM_BUILD_PATH" ]]; then
-    chown -R $CHOWN_FILES_TO $CUSTOM_BUILD_PATH
-  fi
+  chown -R $CHOWN_FILES_TO $BUILD_PATH_FULL
+  chown -R $CHOWN_FILES_TO $UNITY_PROJECT_PATH
 fi
 
 #
--- a/dist/steps/build.sh
+++ b/dist/steps/build.sh
@@ -143,6 +143,19 @@ if [[ "$BUILD_TARGET" == "StandaloneOSX" ]]; then
   chmod +x $ADD_PERMISSIONS_PATH
 fi
 
+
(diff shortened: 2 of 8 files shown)


webbertakken:
Just to confirm. This is an alternative solution to making the images run as a user (in a sense that one does not need the other), correct?

dist/steps/build.sh
```diff
@@ -143,6 +143,19 @@ if [[ "$BUILD_TARGET" == "StandaloneOSX" ]]; then
...
+    chown -R $CHOWN_FILES_TO $BUILD_PATH_FULL
+  fi
+  if [[ -n "$UNITY_PROJECT_PATH" ]]; then
+    chown -R $CHOWN_FILES_TO $UNITY_PROJECT_PATH
+  fi
+  if [[ -n "$CUSTOM_BUILD_PATH" ]]; then
+    chown -R $CHOWN_FILES_TO $CUSTOM_BUILD_PATH
+  fi
+fi
+
```
webbertakken, line 158: this can be simplified to 

```suggestion
if [[ -n "$CHOWN_FILES_TO" ]]; then
  chown -R $CHOWN_FILES_TO $BUILD_PATH_FULL
fi

```

Bradshaw:
> Just to confirm. This is an alternative solution to making the images run as a user (in a sense that one does not need the other), correct?

Right, this is closer to a workaround than a full-on fix, and does not need any changes to be made to the images

That said, I think it's worthwhile to do anyway. When running an image as a user, it needs to be a user that has been defined in the image (AFAIK), so someone wanting to have files owned by a given user on the host machine would need to build custom images.

action.yml
```diff
@@ -106,6 +106,10 @@ inputs:
 
       Parameters must start with a hyphen (-) and may be followed by a value (without hyphen).
       Parameters without a value will be considered booleans (with a value of true).
+  chownFilesTo:
+    required: false
+    default: ''
+    description: After building, change the ownership of the created files to this user
```
webbertakken, line 112:
```suggestion
    description: 'User to give ownership of the resulting build artifacts'
```

dist/steps/build.sh
Bradshaw: The build path may be somewhere completely different to the project path, and the project path has Library and Logs directories created and owned by root, as well as the UnityBuilderAction files copied into the project 

Chowning $CUSTOM_BUILD_PATH might be redundant though indeed

webbertakken: Right unity project path will also be needed. Editing suggestion

webbertakken:
Will merge if you can confirm that this works :)

webbertakken (approved):

Bradshaw:
The config
Screenshot 2021-05-02 at 01 13 23

The files
Screenshot 2021-05-02 at 01 13 37

The developer
thumbs

PR #264, opened 2021-05-23
Remote builder refactoring

frostebite:
Refactors the existing remote builder code.

Implements basic steam deployment from a remote builder by an environment variable. This feature should be considered experimental for now, docs will come later.
`STEAM_DEPLOY`

Implements an environment variable that can be used to purge the remote builder cache.
`PURGE_REMOTE_BUILDER_CACHE`

Implments an environment variable to override the cache key, default cache key also now includes full branch name.
`REMOTE_BUILDER_CACHE`


src/model/remote-builder/remote-builder.ts
```diff
@@ -0,0 +1,422 @@
...
+        await RemoteBuilder.DeployToSteam(buildUid, buildParameters, defaultSecretsArray);
+      }
+    } catch (error) {
+      core.setFailed(error);
+      core.error(error);
+    }
+  }
+  // static async ProvisionPlatformResources() {
+  //   await AWSBuildPlatform.setupPlatformResources();
+  // }
```
webbertakken, line 50: Probably best to remove or add a comment above it what the plan for this snippet is.

src/model/remote-builder/remote-builder.ts
```diff
@@ -0,0 +1,422 @@
...
+          echo ' '
+          echo 'Repo:'
+          ls /${efsDirectoryName}/${buildUid}/${repositoryDirectoryName}/
+          echo ' '
+          echo 'Project:'
+          ls /${efsDirectoryName}/${buildUid}/${repositoryDirectoryName}/${buildParameters.projectPath}
+          echo ' '
+          echo 'Library:'
+          ls /${efsDirectoryName}/${buildUid}/${repositoryDirectoryName}/${buildParameters.projectPath}/Library/
+          echo ' '
```
webbertakken, line 120: Is it possible to use empty lines above the comments? or does that terminate execution?

It would make it easier to read I think.

webbertakken (approved):

webbertakken:
Amazing work as always @frostebite!

src/model/remote-builder/remote-builder.ts
frostebite: empty lines replacing echo would result in no line separation in the logs.

Do you prefer `echo`?

PR #297, opened 2021-11-15
Improve androidVersionCode parsing and application

davidmfinol:
#### Changes

- Increase limit per https://github.com/game-ci/documentation/pull/222#issuecomment-936099761
- Don't apply androidVersionCode if Versioning strategy is set to none.

#### Checklist

### DIFF
--- a/dist/default-build-script/Assets/Editor/UnityBuilderAction/Versioning/VersionApplicator.cs
+++ b/dist/default-build-script/Assets/Editor/UnityBuilderAction/Versioning/VersionApplicator.cs
@@ -15,7 +15,12 @@ namespace UnityBuilderAction.Versioning
     }
 
     public static void SetAndroidVersionCode(string androidVersionCode) {
-      PlayerSettings.Android.bundleVersionCode = Int32.Parse(androidVersionCode);
+      int bundleVersionCode = Int32.Parse(androidVersionCode);
+      if (bundleVersionCode <= 0) {
+        return;
+      }
+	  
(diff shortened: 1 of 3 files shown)


dist/default-build-script/Assets/Editor/UnityBuilderAction/Versioning/VersionApplicator.cs
```diff
@@ -15,7 +15,12 @@ public static void SetVersion(string version)
     }
 
     public static void SetAndroidVersionCode(string androidVersionCode) {
-      PlayerSettings.Android.bundleVersionCode = Int32.Parse(androidVersionCode);
+      int bundleVersionCode = Int32.Parse(androidVersionCode);
+      if (bundleVersionCode <= 0) {
+        return;
+      }
```
webbertakken, line 21: I can't think of any case, but are you sure this won't break anyone's workflow?

webbertakken (approved):

dist/default-build-script/Assets/Editor/UnityBuilderAction/Versioning/VersionApplicator.cs
davidmfinol: Yes, the [version code must be a positive integer >=0](https://developer.android.com/studio/publish/versioning) 

If the user is building for the very first time, they should start with `androidVersionCode = 1`, so if they try to set `androidVersionCode = 0`, then the build will use Unity's default for the version code (ie the value inside their `ProjectSettings` folder).

PR #298, opened 2021-11-18
Add androidTargetSdkVersion as an option to update the Android SDK

davidmfinol:
#### Changes

- Add androidTargetSdkVersion as an option to update the Android SDK

#### Checklist

### DIFF
--- a/src/model/android-versioning.test.ts
+++ b/src/model/android-versioning.test.ts
@@ -24,4 +24,14 @@ describe('Android Versioning', () => {
       expect(AndroidVersioning.determineVersionCode('1.2.3', 2)).toBe(2);
     });
   });
+
+  describe('determineSdkManagerParameters', () => {
+    it('defaults to blank', () => {
+      expect(AndroidVersioning.determineSdkManagerParameters('AndroidApiLevelAuto')).toBe('');
+    });
+
+    it('use specified api level', () => {
+      expect(AndroidVersioning.determineSdkManagerParameters('AndroidApiLevel30')).toBe('platforms;android-30');
+    });
+  });
 });
--- a/action.yml
+++ b/action.yml
@@ -110,6 +110,11 @@ inputs:
     required: false
     default: ''
     description: 'SSH Agent path to forward to the container'
+  gitPrivateToken:
+    required: false
+    default: ''
+    description: >
+      Github private token to pull from github
   chownFilesTo:
     required: false
     default: ''
--- a/action.yml
+++ b/action.yml
@@ -100,7 +100,7 @@ inputs:
     description: 'The android keyaliasPass'
   androidTargetSdkVersion:
     required: false
-    default: 'AndroidApiLevelAuto'
+    default: ''
     description: 'The android target API level.'
   customParameters:
     required: false
--- a/action.yml
+++ b/action.yml
@@ -98,6 +98,10 @@ inputs:
     required: false
     default: ''
     description: 'The android keyaliasPass'
+  androidTargetSdkVersion:
+    required: false
+    default: 'AndroidApiLevelAuto'
+    description: 'The android target API level.'
   customParameters:
     required: false
     default: ''
--- a/dist/default-build-script/Assets/Editor/UnityBuilderAction/Input/AndroidSettings.cs
+++ b/dist/default-build-script/Assets/Editor/UnityBuilderAction/Input/AndroidSettings.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using UnityEditor;
 
 namespace UnityBuilderAction.Input
--- a/dist/default-build-script/Assets/Editor/UnityBuilderAction/Input/AndroidSettings.cs
+++ b/dist/default-build-script/Assets/Editor/UnityBuilderAction/Input/AndroidSettings.cs
@@ -16,6 +16,8 @@ namespace UnityBuilderAction.Input
         PlayerSettings.Android.keyaliasName = keyaliasName;
       if (options.TryGetValue("androidKeyaliasPass", out string keyaliasPass) && !string.IsNullOrEmpty(keyaliasPass))
         PlayerSettings.Android.keyaliasPass = keyaliasPass;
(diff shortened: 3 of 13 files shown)


davidmfinol:
A run where I don't set the `androidTargetSdkVersion`: https://github.com/finol-digital/Card-Game-Simulator/actions/runs/1474689325
And one where I do set the `androidTargetSdkVersion` to level 30: https://github.com/finol-digital/Card-Game-Simulator/actions/runs/1474802849

Let me know if there any additional tests I should run, or if there's any other concerns/questions.

davidmfinol:
I'm getting this error:
```
ERROR: JAVA_HOME is set to an invalid directory: /opt/unity/Editor/Data/PlaybackEngines/AndroidPlayer/Tools/OpenJDK/Linux

Please set the JAVA_HOME variable in your environment to match the
location of your Java installation.
```

I tried to set the JAVA_HOME the way that it's done here: https://github.com/game-ci/docker/blob/main/editor/Dockerfile#L100
Then again, I shouldn't even need to set JAVA_HOME since it was already done there, right?
Maybe you know what I am doing wrong, @GabLeRoux ?

src/model/android-versioning.test.ts
```diff
@@ -28,4 +28,14 @@ describe('Android Versioning', () => {
       expect(AndroidVersioning.determineVersionCode('1.2.3', 2)).toBe(2);
     });
   });
+
+  describe('determineSdkManagerParameters', () => {
+    it('defaults to blank', () => {
+      expect(AndroidVersioning.determineSdkManagerParameters('AndroidApiLevelAuto')).toBe('');
+    });
+
+    it('use specified api level', () => {
```
webbertakken, line 37:
```suggestion
    it('uses the specified api level', () => {
```

webbertakken (approved):
Amazing work, looks very clean and I like that you're only exposing one parameter and determine the rest in `BuildParameters.create`.

webbertakken:
re @davidmfinol:

> I'm getting this error:
> 
> ```
> ERROR: JAVA_HOME is set to an invalid directory: /opt/unity/Editor/Data/PlaybackEngines/AndroidPlayer/Tools/OpenJDK/Linux

Have you tried using `$UNITY_PATH`?

davidmfinol:
Same issue when I use `UNITY_PATH`: https://github.com/finol-digital/Card-Game-Simulator/runs/4258191146?check_suite_focus=true#step:6:333

jsteinich:
That referenced line in the editor docker file is actually only used by older versions of Unity. Lines 155-158 are what is used on newer versions (certainly a bit more convoluted).

One thing I've ran into is depending on how the underlying container is accessed impacts whether or not the underlying environment variables have been exported (set in a unity wrapper script or ~/.bashrc only).

Ideally it would be worth figuring out why neither the wrapper script or ~/.bashrc is being invoked to currently set the variables.

We've used an ugly workaround by invoking the sdkmanager as follows:
`"$(awk -F'=' '/ANDROID_HOME=/{print $2}' /root/.bashrc)/tools/bin/sdkmanager"`.
Our usage has a different version of java set for JAVA_HOME, but the same idea could be applied to JAVA_HOME.
Note: We're using that workaround because we are actually building our own editor images on top of the base ones with extra tools added.

davidmfinol:
> `"$(awk -F'=' '/ANDROID_HOME=/{print $2}' /root/.bashrc)/tools/bin/sdkmanager"`.
> Our usage has a different version of java set for JAVA_HOME, but the same idea could be applied to JAVA_HOME.

I'm not skilled with bash, so please help me understand.
It looks like `awk -F'=' '/ANDROID_HOME=/{print $2}' /root/.bashrc` will read the value of ANDROID_HOME from `/root/.bashrc`, right? 
Is the JAVA_HOME value also inside `/root/.bashrc`?
Could I do something like `export JAVA_HOME="$(awk -F'=' '/JAVA_HOME=/{print $2}' /root/.bashrc)"`?

jsteinich:
> It looks like `awk -F'=' '/ANDROID_HOME=/{print $2}' /root/.bashrc` will read the value of ANDROID_HOME from `/root/.bashrc`, right?

That is correct.



> Is the JAVA_HOME value also inside `/root/.bashrc`?
> Could I do something like `export JAVA_HOME="$(awk -F'=' '/JAVA_HOME=/{print $2}' /root/.bashrc)"`?

Yep.



> Ideally it would be worth figuring out why neither the wrapper script or ~/.bashrc is being invoked to currently set the variables.

Since the android sdk step is done prior to invoking the `unity-editor` wrapper script, only `/root/.bashrc` would have a chance to run; however, the docker container is ran with `--volume /home/runner/work/_temp/_github_home:/root`, so that file is effectively clobbered.

davidmfinol:
I get this error: `awk: cannot open /root/.bashrc (No such file or directory)`

jsteinich:
> I get this error: `awk: cannot open /root/.bashrc (No such file or directory)`

Ah, yes. It's failing for the same reason that `/root/.bashrc` is failing in the 1st place.
Looks like the updated docker file also writes to `/usr/bin/unity-editor.d/`, though the exact file is named differently depending on Unity version.

I just did a quick test, and replacing `export JAVA_HOME="$(awk -F'=' '/JAVA_HOME=/{print $2}' /root/.bashrc)"` with `export JAVA_HOME="$(awk -F'=' '/JAVA_HOME=/{print $2}' /usr/bin/unity-editor.d/*)"` is working for me in the unity 2020.3.23f1-android-0.15.0 image. (Need to do a similar update for android home)

davidmfinol:
Great, that works!
Here's my passing test run: https://github.com/finol-digital/Card-Game-Simulator/actions/runs/1497846444

If no one has any concerns, I'll merge this tomorrow.

GabLeRoux (approved):

webbertakken (approved):

trumpets:
Don't wanna review an old PR, but I just tried using the new param and it seems the actions haven't been updated since this PR was merged. 

Any hints on what the process could be here? Is is something an outsider can contribute or a specific process needs to be triggered?

trumpets:
Thank you @webbertakken ! I didn't know I could configure the action to use another version. Very cool!

PR #305, opened 2021-12-19
Initial Support for Windows Based Builds

AndrewKahr:
Implemented logic for windows-based docker builds. Moved dockerfiles and scripts to platform specific folders.

#### Changes

- Added win32 as a supported base platform
- Added logic for pulling windows based docker images
- Added logic for mounting windows based file systems to docker
- Added checks for required windows components on the runner
- Moved dockerfiles, entrypoint, and steps scripts to platform-based directories
- Added dockerfile, entrypoint, and steps scripts for windows
- Modified default build script to check for an android build before attempting to access an Android version variable that may not exist
- Added various error messages for invalid configurations, ie: windows based runner trying to run a windows image on a version below 2019.3, missing required environment secrets for windows, missing required build tools on the host runner that are needed within the container.

#### Checklist


src/model/image-tag.test.ts
```diff
@@ -64,13 +70,27 @@ describe('ImageTag', () => {
...
       const image = new ImageTag({ version: '2019.2.11f1', platform: 'WebGL' });
 
-      expect(image.toString()).toStrictEqual(`${defaults.image}:2019.2.11f1-webgl-0`);
+      switch (process.platform) {
+        case 'win32':
+          expect(image.toString()).toStrictEqual(`${defaults.image}:windows-2019.2.11f1-webgl-0`);
+          break;
+        case 'linux':
+          expect(image.toString()).toStrictEqual(`${defaults.image}:2019.2.11f1-webgl-0`);
+          break;
```
webbertakken, line 79: Just a note to convey context:

This is where it becomes tricky. Over time each target platform may dictate which base platform works best for it. Perhaps we can use some magic by then to reflect this in user's workflows.

This means that over time we'll have to become very specific about which target platform uses which base platform. At that point this would likely become `${defaults.image}:ubuntu-2019.2.11f1-webgl-0`, with the `ubuntu` part to be specific.

For now we'll just assume that the default is always ubuntu. Just starting the thought engine on how to architect it over time.

src/model/image-tag.ts
```diff
@@ -35,25 +35,61 @@ class ImageTag {
...
+        // Can only build windows-il2cpp on a windows based system
+        if (process.platform === 'win32') {
+          // Unity versions before 2019.3 do not support il2cpp
+          if (major >= 2020 || (major === 2019 && minor >= 3)) {
+            return windowsIl2cpp;
+          } else {
+            throw new Error(`Windows-based builds are only supported on 2019.3.X+ versions of Unity.
+                             If you are trying to build for windows-mono, please use a Linux based OS.`);
+          }
+        }
```
webbertakken, line 92: This looks great for the platform selection part. Love it!

test-project-windows/.gitattributes
```diff
@@ -0,0 +1,86 @@
+﻿* text=auto
+
+# Unity files
```
webbertakken, line 3: Do we need another test project? Ideally we'd reuse existing projects where possible.

We can merge with it but it's definitely something to keep in mind.

test-project-windows/Packages/manifest.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "com.unity.ide.visualstudio": "2.0.12"
+  }
```
webbertakken, line 4: I'm not an expert at Unity, could you please explain why this would be needed as a dependency?

webbertakken (approved):
Great work! LGTM

Generally approved, but lets try a test run (i.e. `with: game-ci/unity-builder@<specific-sha>`) before merging it in case we're missing small bits.

To test it we first need to publish the images, for which I'm going to make some changes in https://github.com/game-ci/versioning-backend. I'll try to do this very soon.

test-project-windows/Packages/manifest.json
AndrewKahr: Gonna be honest, I'm not too sure either. I think that got auto added when I opened the project to configure it. We might be able to just remove it.

test-project-windows/.gitattributes
AndrewKahr: Reason I made a copy of the project was because I needed to set the scripting backend to il2cpp. I'm not sure if we can do that from the cli since the target is just StandaloneWindows with no mention of the scripting backend. Open to suggestions on how to get it to test with the right backend without a separate project.

timcassell: That way I do it in my own project is I overwrite the `scriptingBackend` in the `ProjectSettings.asset` before running the build. But that also requires me to include `allowDirtyBuild: true`, so not sure if that makes sense here.

webbertakken: Just for testing the integrity of the action it might be worth it.

We do have some examples of where we already do this. ([example 1](https://github.com/game-ci/docker/blob/c82b022be4c88573b2f4d5f4aece4945e0824b30/.github/workflows/test.yml#L208), [example 2](https://github.com/game-ci/docker/blob/c82b022be4c88573b2f4d5f4aece4945e0824b30/.github/workflows/test.yml#L230))

AndrewKahr: Sounds good. I'll see about getting something similar implemented. Thanks for pointing me in the right direction 🙂

test-project-windows/Packages/manifest.json
davidmfinol: It's used to enable some integrations that Visual Studio has with Unity (like connecting to the Unity process, so you are able to setup breakpoints for debugging). I'd recommend either a) removing it or b) adding for all other IDEs like "com.unity.ide.rider" and "com.unity.ide.vscode". Probably a) is better.

dist/platforms/windows/Dockerfile
```diff
@@ -0,0 +1,18 @@
+ARG IMAGE
+FROM $IMAGE
+
+LABEL "com.github.actions.name"="Unity - Builder"
+LABEL "com.github.actions.description"="Build Unity projects for different platforms."
+LABEL "com.github.actions.icon"="box"
+LABEL "com.github.actions.color"="gray-dark"
+
+LABEL "repository"="http://github.com/webbertakken/unity-actions"
```
davidmfinol, line 9: We should probably take this opportunity to change this to `http://github.com/game-ci/unity-actions`

davidmfinol (approved):

webbertakken:
Note that for me to be able to merge it, you'll have to either merge main into this branch or rebase the branch and force-push it.

Will look into building the images for testing this tomorrow.

AndrewKahr:
@webbertakken How do I fix merge conflicts for the index.js.map file? Do I just remove it and it gets regenerated on commit?

AndrewKahr:
Nvm I added the -X theirs flag and it did the merge

webbertakken:
> @webbertakken How do I fix merge conflicts for the index.js.map file? Do I just remove it and it gets regenerated on commit?

Indeed. It regenerates on a pre-commit hook. or you can `yarn build` manually.

AndrewKahr:
There's now a compatibility check on tests that requires the base system to be linux or it throws an error in actions.test.ts. But when committing on windows it still runs that check and fails which cancels the commit. How do we want to proceed on this? Should windows be added as a valid platform?

webbertakken:
Right. Feel free to delete the test or have it accept windows.

AndrewKahr:
I just realized I read the test wrong. I thought it was saying the tests only run on linux. I've added windows as an allowed platform for the test now.

src/model/image-tag.ts
```diff
@@ -35,25 +35,61 @@ class ImageTag {
       webgl: 'webgl',
       mac: 'mac-mono',
       windows: 'windows-mono',
+      windowsIl2cpp: 'windows-il2cpp',
+      wsaplayer: 'universal-windows-platform',
```
webbertakken, line 39: could you camelcase `wsaPlayer` here please? I'm working on syncing versioning backend with your changes.

AndrewKahr: Sure, should it be WSAPlayer since wsa is an acronym or just wsaPlayer?

webbertakken: We use lowercase acronyms in all my projects including all GameCI repos.

See also most accepted answer here https://stackoverflow.com/questions/15526107/acronyms-in-camelcase

webbertakken:
Here's part 1 https://github.com/game-ci/versioning-backend/pull/36

AndrewKahr:
What's the next step? Is the backend already up and now we try building an image and a project?

webbertakken:
The backend is what you see spamming in #news on discord every time there's a new Unity editor version or docker repository version out.

As for the steps, I'm thinking this:

1. Merge https://github.com/game-ci/versioning-backend/pull/36
1. Merge https://github.com/game-ci/docker/pull/141
1. Continue with the rest of the code that's needed to support windows builds on versioning backend
1. Run all windows builds over the course of 36 hours

During the last step we can simultaneously 
- debug failing windows builds, as well as 
- start testing windows baseed builds on unity-builder using the last commit from this PR and eventually merge this PR

wdyt?

AndrewKahr:
Sounds good to me. Let's make it happen!

src/model/docker.ts
```diff
@@ -111,9 +110,7 @@ class Docker {
                              steps and set the UNITY_LICENSE GitHub secret or enter a Unity
                              serial number inside the UNITY_SERIAL GitHub secret.`);
           }
-          unitySerial = this.getSerialFromLicenseFile(process.env.UNITY_LICENSE);
-        } else {
-          unitySerial = process.env.UNITY_SERIAL!;
+          process.env.UNITY_SERIAL = this.getSerialFromLicenseFile(process.env.UNITY_LICENSE);
```
webbertakken, line 113: Note that it is not good practice to assign to environment from CI. GitHub has other ways to make it as masked.

AndrewKahr: For some reason GitHub wasn't masking secrets in the calls. Even when I added the serial as a secret explicitly. Perhaps I was doing it wrong? This is more a bandaid solution though and am open to ways to fix it. It would be best to keep the serial from being required in the secrets as people with free licenses will need to do extra work to get the serial in there.

webbertakken: Here's the [official docs](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#masking-a-value-in-log) on how to achieve want you wanna do. 


Out of scope for this PR though. We need to absolutely keep things manageable when it comes to bugs.

AndrewKahr: Just added setSecret to the latest commit and it seems to be working. Thanks for pointing me in the right direction!

dist/default-build-script/Assets/Editor/UnityBuilderAction/Builder.cs
```diff
@@ -39,12 +39,14 @@ public static void BuildProject()
 
       // Set version for this build
       VersionApplicator.SetVersion(options["buildVersion"]);
-      VersionApplicator.SetAndroidVersionCode(options["androidVersionCode"]);
       
       // Apply Android settings
       if (buildPlayerOptions.target == BuildTarget.Android)
+      {
+        VersionApplicator.SetAndroidVersionCode(options["androidVersionCode"]);
```
webbertakken, line 46: @davidmfinol does this look correct to you too?

davidmfinol: Yep! I think it makes sense to move this line so it's only applied when it's actually needed.

dist/platforms/windows/steps/build.ps1
```diff
@@ -0,0 +1,147 @@
...
+if ($Env:BUILD_EXIT_CODE -eq 0) 
+{
+    Write-Output "Build Succeeded!"
+} else 
+{
+    Write-Output "$('Build failed, with exit code ')$($Env:BUILD_EXIT_CODE)$('"')"
+}
+
+# TODO: Determine if we need to set permissions on any files
+
```
webbertakken, line 135: I believe this todo is now solved as everything worked?

AndrewKahr: Everything built and ran so it looks good on my end. What did not setting permissions do on the Linux side? Did things fail to build/run?

src/model/action.ts
```diff
@@ -30,7 +30,16 @@ class Action {
   }
 
   static get dockerfile() {
-    return `${Action.actionFolder}/Dockerfile`;
+    const currentPlatform = process.platform;
+    switch(currentPlatform)
+    {
```
webbertakken, line 35:
```suggestion
    switch (currentPlatform) {
```

src/model/docker.ts
```diff
@@ -1,5 +1,7 @@
+import * as core from '@actions/core';
 import { exec } from '@actions/exec';
 import ImageTag from './image-tag';
+const fs = require('fs');
```
webbertakken, line 4:
```suggestion
import fs from 'fs';
```

.github/workflows/build-tests.yml
```diff
@@ -68,6 +68,68 @@ jobs:
...
+    runs-on: windows-2019
+    strategy:
+      fail-fast: false
+      matrix:
+        projectPath:
+          - test-project
+        unityVersion:
+          - 2020.3.24f1
+        targetPlatform:
+          - StandaloneWindows64 # Build a Windows 64-bit standalone.
```
davidmfinol, line 86: Can we add `StandaloneWindows` here as well? Just as an additional sanity check, since I think 32-bit windows build should obviously be able to be created with a windows runner:
```suggestion
          - StandaloneWindows64 # Build a Windows 64-bit standalone.
          - StandaloneWindows # Build a Windows 32-bit standalone.
```

src/model/docker.ts
```diff
@@ -95,7 +99,173 @@ class Docker {
...
+
+        if (!(process.env.UNITY_EMAIL && process.env.UNITY_PASSWORD)) {
+          throw new Error(`Unity email and password must be set for Windows based builds to
+                           authenticate the license. Make sure to set them inside UNITY_EMAIL
+                           and UNITY_PASSWORD in Github Secrets and pass them into the environment.`);
+        }
+
+        await this.setupWindowsRun(platform);
+
+        this.validateWindowsPlatformRequirements(platform);
```
webbertakken, line 129: It's fine to do this in a follow-up as well, but this is what the `input` and `buildParameters` models are for. We shouldn't inline some last minute adjustments right before the build command.

The docker model class should only know about how to run the docker image, nothing more.

src/model/docker.ts
```diff
@@ -95,7 +99,173 @@ class Docker {
...
+
+        const windowsRunCommand = `docker run \
+        --workdir c:/github/workspace \
+        --rm \
+        --env UNITY_LICENSE \
+        --env UNITY_LICENSE_FILE \
+        --env UNITY_EMAIL \
+        --env UNITY_PASSWORD \
+        --env UNITY_SERIAL="${unitySerial}" \
+        --env UNITY_VERSION="${version}" \
```
webbertakken, line 139: Looks like we do in fact not need a switch (baseOs) here. The parameters should be passed taking the baseOs in mind.

That and other checks currently happen in the factory method (`create`) of the build-parameters model.

We should either make the OS-modifications to the parameters inside `build-parameters` or create another layer in between for OS specific stuff only. 

Personally I think 
- `unitySerial` can be parsed inside`build-parameters`
- `--workdir c:/github/workspace \` can be specific to the `docker` class, but like `getWorkDirParam(buildPlatform)` or something similar
- `await this.setupWindowsRun(platform);` can also be handled in `docker` class but like `setupBuildPlatform(buildPlatform)`

Note that these would not be functional changes, just moving around some code to make it more manageable.

webbertakken (approved):
Looks good overall. 

Small point on code organisation, but it's fine to pick that up after merging as well if you prefer.

dist/platforms/windows/steps/build.ps1
webbertakken: I see the build wasn't started yet. I've just approved the run. It always does that for first time contributors (for each repo)

src/model/image-tag.ts
davidmfinol: Since this code is duplicated for both the `Platform.types.StandaloneWindows` and `Platform.types.StandaloneWindows64` cases, what do you think of simply doing a fall through? Some people don't like fall throughs (some people think switch statements should always be avoided), but I think this is the exact situation where a switch fall through is good since it removes duplicate code:
```suggestion
      case Platform.types.StandaloneWindows64:
        // Can only build windows-il2cpp on a windows based system
        if (process.platform === 'win32') {
          // Unity versions before 2019.3 do not support il2cpp
          if (major >= 2020 || (major === 2019 && minor >= 3)) {
            return windowsIl2cpp;
          } else {
            throw new Error(`Windows-based builds are only supported on 2019.3.X+ versions of Unity.
                             If you are trying to build for windows-mono, please use a Linux based OS.`);
          }
        }
```

davidmfinol (approved):
Great work! Just 2 small suggestions from me

webbertakken:
@davidmfinol Is the code from both branches exactly the same? In that case I also agree. If not the quality checks will not allow this because the complexity rises harder than code deduplication offers.

davidmfinol:
> @davidmfinol Is the code from both branches exactly the same? looks like right? In that case I also agree. If not the quality checks will not allow this because the complexity rises harder than code deduplication offers.

Yep, its the exact same. If it weren't I would have recommended extracting the duplicate code into a function, and calling that function from both branches.

AndrewKahr:
I did try to do fall through cases before but the precommit hook fails the build as the standard does not allow for it.

webbertakken:
> I did try to do fall through cases before but the precommit hook fails the build as the standard does not allow for it.

I just tested this. It allows it, but only if the cases are the same.

For example this shows an error:

image

But this does not

image

src/model/docker.ts
AndrewKahr: The big change on the windows branch is passing in all of the windows specific directories like Visual Studio, win10sdk registry keys, etc. How do we get that through without a branch?

webbertakken: My suggestion would be to use a branch, but more localised. 

For example to a method that returns a string with the platform specific only.

```
docker.run(`
   same
   same
   same
   ${platformSpecificDirectories(baseOs)}
`)
```

AndrewKahr:
I now have the unity serial code in build parameters but now tests don't pass because there is no license for it to pull a serial from thus causing it to throw an exception (which is expected) but that means none of the tests can run. How should I fix the test?

webbertakken:
> I now have the unity serial code in build parameters but now tests don't pass because there is no license for it to pull a serial from thus causing it to throw an exception (which is expected) but that means none of the tests can run. How should I fix the test? Can I force environment variables somehow and put the license that is used in the workflows?

If I understand your comment correctly we probably have to do this: https://stackoverflow.com/a/58953365/3593896

src/model/docker.ts
```diff
@@ -45,227 +44,103 @@ class Docker {
...
-        this.generateWinSDKRegKeys(silent);
-        break;
-      case 'StandaloneWindows64':
-        this.generateWinSDKRegKeys(silent);
-        break;
-      case 'WSAPlayer':
-        this.generateWinSDKRegKeys(silent);
-        break;
-    }
+    const baseOSSpecificArgs = this.getBaseOSSpecificArgs(process.platform, parameters);
```
webbertakken, line 47: Reminder for camelCase acronyms: `baseOsSpecificArguments` :)

PR #310, opened 2021-12-28
Cloud Runner v2

frostebite:
- LFS caching
- Library caching
- improved simple K8s support
- improved refactoring of AWS support
- configurable pre jobs and post jobs
- experimental hook to inject configurable bash commands at the end or start of any job
- initial test coverage for cloud runner
- command-line mode support

PR #326, opened 2022-01-26
Initial Support for MacOS IL2CPP Builds

AndrewKahr:
#### Changes

- Ported Linux scripts for use on macOS runner
- Added bash script to install unity hub/editor on the macOS runner in lieu of docker support
- Updated action to support macOS and call appropriate bootstrap script instead of docker container
- Adds [Unity-Changeset](https://github.com/mob-sakai/unity-changeset) to yarn dependencies for getting the Unity changeset given the version
- Fixes building library during activation. Changes backported to Windows as well.

#### Checklist


dist/platforms/mac/entrypoint.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+#
+# Run steps
+#
+source $ACTION_FOLDER/platforms/mac/steps/setup.sh
```
webbertakken, line 6: Perhaps this setup can be done from the TypeScript part instead so we can completely reuse the existing entrypoint and related scripts from the linux variant.

Reasons:
- These files have been empirically improved upon for 3 years, diverging them into mutliple files with 95%  the same content would make our implementation less cohesive.
- These files are intended to be docker runtime, not build time - if anything we should be adding a layer in between that installs it, or mount the installation from the runners host.

dist/platforms/mac/steps/activate.sh
```diff
@@ -2,6 +2,9 @@
 
 echo "Requesting activation"
 
+# Need this because it tries to initialize the library when activating
+UNITY_PROJECT_PATH="$GITHUB_WORKSPACE/$PROJECT_PATH"
```
webbertakken, line 6: Not sure if you got that part, but we already work around this in the Ubuntu activate file.

Passing a projectPath did not work as expected because:
- if you pass an empty path, it won't recognise it as a project
- if you pass the real project, it'll initialise library during this step
- if you pass no parameter and you're in the default folder while the project is checked out in that folder too, it'll also initialise the library during this step

The only way we managed to make this work consistently was to
- move to a different path
- use no parameter so it uses the default behaviour (which is creating a new project which doesn't need initialising) in that different path
- switch back to the previous path.

webbertakken:
What is the status of this? Will you let us know when it's ready for review?

Did you get it to work?

AndrewKahr:
@webbertakken Apologies, I forgot to re-request a review. It is ready for a new review. On my fork it successfully built the test-project for all platforms.

I'm trying to now merge upstream into my branch and am having some issues. When committing, jest fails to find tests, and the commit is canceled. When I run yarn test several tests fail for system.test.ts where it is unable to run commands. I also get a lot of these errors: `Command failed: gh auth status -t
    'gh' is not recognized as an internal or external command,
    operable program or batch file.`
I reran yarn and yarn build and this continues to happen. Is there an issue with the tests right now or am I doing something wrong?

webbertakken:
Oh wow it's already working! Nice job.

What you could try is run `yarn` then `git commit -am "chore: update dist files" --no-verify` to get the commit in first, and push it.

After that you can run `yarn test` and `yarn build` again I think.

.github/workflows/mac-build-tests.yml
```diff
@@ -0,0 +1,74 @@
...
+
+      ###########################
+      #          Cache          #
+      ###########################
+      - uses: actions/cache@v2
+        with:
+          path: ${{ matrix.projectPath }}/Library
+          key: Library-${{ matrix.projectPath }}-windows-${{ matrix.targetPlatform }}
+          restore-keys: |
+            Library-${{ matrix.projectPath }}-windows-
```
webbertakken, line 41:
```suggestion
          key: Library-${{ matrix.projectPath }}-macos-${{ matrix.targetPlatform }}
          restore-keys: |
            Library-${{ matrix.projectPath }}-macos-
```

.github/workflows/mac-build-tests.yml
```diff
@@ -0,0 +1,74 @@
...
+          UNITY_EMAIL: ${{ secrets.UNITY_EMAIL }}
+          UNITY_PASSWORD: ${{ secrets.UNITY_PASSWORD }}
+          UNITY_SERIAL: ${{ secrets.UNITY_SERIAL }}
+        with:
+          projectPath: ${{ matrix.projectPath }}
+          unityVersion: ${{ matrix.unityVersion }}
+          targetPlatform: ${{ matrix.targetPlatform }}
+          customParameters: -profile SomeProfile -someBoolean -someValue exampleValue
+          allowDirtyBuild: true
+      # We use dirty build because we are replacing the default project settings file above
```
webbertakken, line 65:
```suggestion
          # We use dirty build because we are replacing the default project settings file above
          allowDirtyBuild: true
```

dist/BlankProject/Packages/packages-lock.json
```diff
@@ -0,0 +1,4 @@
+{
+  "dependencies": {
+    }
+}
```
webbertakken, line 4:
```suggestion
{
  "dependencies": {
  }
}
```

dist/BlankProject/Packages/manifest.json
```diff
@@ -0,0 +1,4 @@
+{
+  "dependencies": {
+  }
+}
```
webbertakken, line 4: Pretty cool that this works :)

src/model/platform-setup/setup-mac.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    await SetupMac.setEnvironmentVariables(buildParameters, actionFolder);
+  }
+
+  private static async installUnityHub(silent = false) {
+    const command = 'brew install unity-hub';
+    if (!fs.existsSync(this.unityHubPath)) {
+      //Ignoring return code because the log seems to overflow the internal buffer which triggers
+      //a false error
```
webbertakken, line 25:
```suggestion
      // Ignoring return code because the log seems to overflow the internal buffer which triggers
      // a false error
```

src/model/platform-setup/setup-mac.ts
```diff
@@ -0,0 +1,75 @@
...
+    //a false error
+    const errorCode = await exec(command, undefined, { silent, ignoreReturnCode: true });
+    if (errorCode) {
+      throw new Error(`There was an error installing the Unity Editor. See logs above for details.`);
+    }
+  }
+
+  private static async setEnvironmentVariables(buildParameters: BuildParameters, actionFolder: string) {
+    //Need to set environment variables from here because we execute
+    //the scripts on the host for mac
```
webbertakken, line 51:
```suggestion
    // Ignoring return code because the log seems to overflow the internal buffer which triggers
    // a false error
    const errorCode = await exec(command, undefined, { silent, ignoreReturnCode: true });
    if (errorCode) {
      throw new Error(`There was an error installing the Unity Editor. See logs above for details.`);
    }
  }

  private static async setEnvironmentVariables(buildParameters: BuildParameters, actionFolder: string) {
    // Need to set environment variables from here because we execute
    // the scripts on the host for mac
```

webbertakken (approved):
LGTM

One thing I was wondering about the entrypoint. Ideally the entrypoint (which runs inside the container) would not know about the "action folder" (which is a concept that ideally only exists outside the of the container) but rather just copy in the right files in the dockerfile. But I'm not sure if that would work here?

AndrewKahr:
> One thing I was wondering about the entrypoint. Ideally the entrypoint (which runs inside the container) would not know about the "action folder" (which is a concept that ideally only exists outside the of the container) but rather just copy in the right files in the dockerfile.
Yea so if this were in the container it would be copied to root and be ready to go for the script. I had considered having it copied to the home directory to mimic root but that would require us to write a recursive folder copy function in ts unless we want to have an additional bash script. We also still have to modify the scripts so they start with ~/ so I found simply adding an env variable with the base path to be simpler.

AndrewKahr:
> After that you can run `yarn test` and `yarn build` again I think.

Gave it another go, `yarn build` is fine, but `yarn test` still blows up. Can you try running it on your own machine? I'm wondering if it's just my setup is broken or if the tests are broken now.

webbertakken:
> > After that you can run `yarn test` and `yarn build` again I think.
> 
> Gave it another go, `yarn build` is fine, but `yarn test` still blows up. Can you try running it on your own machine? I'm wondering if it's just my setup is broken or if the tests are broken now.

I'm getting 10 failures.

image

I believe they may be related to #310 as I'm getting the same errors on main.

@frostebite I saw you made some changes to tests and also used `gh auth`. Perhaps you know why tests are failing?

webbertakken:
@AndrewKahr note that the workflow is failing because of formatting issues.

AndrewKahr:
@webbertakken Fixed 😃

vinaysshenoy:
Question: Does this mean we will be able to run the Unity build action on a self-hosted M1 Mac Mini machine soon?

webbertakken:
> Question: Does this mean we will be able to run the Unity build action on a self-hosted M1 Mac Mini machine soon?

Kindly open a new issue with a question so that everyone can find it. I think this is useful because it might not involve a one-liner answer.

PR #353, opened 2022-03-04
Cloud Runner v0 - Reliable and trimmed down cloud runner

frostebite:
This pull request implements the minimum features to implement the promise of cloud runner fully either via the default parameters or via the extensible override parameters.

### K8s support
unstable/development->experimental
_You can in theory easily accomplish distributed caching via pull cache override for k8s in this release to make k8s a fully viable workflow for massive projects. Will be experimenting with some multi-shared read/write PVs from Rook in the next PR._

### Cloud Runner General Support
experimental->preview
_AWS implementation out of the box can handle massive projects with sensible parameters._

## Preview phase of cloud runner
If accepted I will begin using tag releases to communicate stable releases for cloud runner. `cloud-runner-develop` will continue as the primary integration branch.

The preview phase of the cloud runner release will represent:
- further feedback gathering
- alignment with any needed refinement for the wider game ci project. 
- any needed steps to simplify and align with beginner-friendly promise of game-ci, including but not limited to documentation improvements.
- garbage collection improvements (guarantees, check active resources)
- Moving away from using the branch as reference for cloud runner. Instead, we will use SHA. Currently, we jump to the latest commit on a given branch.
- Integrated ephemeral runner support to cut-down impact of long run times with cloud runner.

Post preview I will explore warm-starts, diff-based runs, generic dependency startup/cleanup hooks, editor integration, extensible logging, gcp run, testing improvements (including tests for caching hit/miss specific performance tests) extensible logging, benchmarking (large reference project), better artifact handling, more explicit checking of build success to make it convenient to expose results into output and probably more...

When satisfied with the above cloud runner will progress to released/ga.

## Other improvements
There is also a general refactoring to make cloud runner _polished and shiny_. For example:
- Many files and functions were removed (such as cloud runner state, which was simplified into a smaller implementation in cloud runner's main file) as part of a general simplification and cutting down on structures that didn't end up having solid responsibilities.
- All async input has been removed and generally cleaned up to better align with the build parameters pattern.
- Fully realized CLI support
- Override input source - use terminal commands to read any input source to make it easy to specify large values/secrets such as a license or steps YAML
- Hooks - (pre/post) (setup/build) container steps and (pre/post/all) (setup/build) terminal command hooks.
- Combined setup/run cloud runner container to minimize startup time.
- Some important increases in test coverage (quick verification of caching and deep verification of caching)


davidmfinol (approved):

src/model/cli/cli.ts
```diff
@@ -3,55 +3,85 @@ import { BuildParameters, CloudRunner, ImageTag, Input } from '..';
...
-    return options.mode !== undefined && options.mode !== '';
+  public static query(key, alternativeKey) {
+    if (CLI.options && CLI.options[key] !== undefined) {
+      return CLI.options[key];
+    }
+    if (CLI.options && alternativeKey && CLI.options[alternativeKey] !== undefined) {
+      return CLI.options[alternativeKey];
+    }
+    return;
   }
```
webbertakken, line 26: `Cli.query` is very abstract in naming. 

Would it make sense to use a common `options` approach where you have overrides that are passed in and then there are no "alternative keys"?

Furthermore, I'd argue that even something like `getOption` would be quite abstract still, and far away from the domain. Abstract code is both difficult to understand and maintain.

An example that could be a bit closer to the domain could look like the following:

Let's assume `key` that's passed in would be `runner`

```ts
private get isUsingGitHubRunner() {
  return this.options.runner === 'github'
}
```

Subsequent code becomes much easier to read:

```ts
if (this.isUsingGitHubRunner) {
  // something specific to this option
}
```

src/model/cloud-runner/cloud-runner.ts
```diff
@@ -1,69 +1,88 @@
...
+        CloudRunner.CloudRunnerProviderPlatform = new TestCloudRunner();
+        break;
+      case 'local-system':
+        CloudRunner.CloudRunnerProviderPlatform = new LocalCloudRunner();
+        break;
+      case 'local-docker':
+        CloudRunner.CloudRunnerProviderPlatform = new LocalDockerCloudRunner();
         break;
     }
   }
```
webbertakken, line 58: I love that you used a strategy pattern for this. It'll come in handy soon enough with our current plans.

src/model/cloud-runner/error/cloud-runner-error.ts
```diff
@@ -1,16 +1,16 @@
...
 export class CloudRunnerError {
   public static async handleException(error: unknown) {
     CloudRunnerLogger.error(JSON.stringify(error, undefined, 4));
     core.setFailed('Cloud Runner failed');
-    await CloudRunnerState.CloudRunnerProviderPlatform.cleanupSharedResources(
-      CloudRunnerState.buildParams.buildGuid,
-      CloudRunnerState.buildParams,
-      CloudRunnerState.branchName,
-      CloudRunnerState.defaultSecrets,
+    await CloudRunner.CloudRunnerProviderPlatform.cleanupSharedResources(
```
webbertakken, line 9: Perhaps we could rename `CloudRunner.CloudRunnerProviderPlatform` to `CloudRunner.Provider`.

We could go one step further and encapsulate knowledge about the Interface to CloudRunner itself. You can do this by calling `CloudRunner.cleanup()` instead. It would then call `CloudRunner.Provider.cleanup()` which would then know how to clean up it's shared resources.

src/model/cloud-runner/remote-client/caching.test.ts
```diff
@@ -0,0 +1,65 @@
...
+      Input.githubInputEnabled = false;
+      const buildParameter = await BuildParameters.create();
+      CloudRunner.buildParameters = buildParameter;
+      // create test folder
+      const testFolder = path.resolve(__dirname, CLI.options.cacheKey);
+      fs.mkdirSync(testFolder);
+      // crate cache folder
+      const cacheFolder = path.resolve(__dirname, `cache-${CLI.options.cacheKey}`);
+      fs.mkdirSync(cacheFolder);
+      // add test has file to test folders
```
webbertakken, line 41: I've worked with `fs` a few times in the past and it looks like `fs.promises` is a very good and modern core library of node. Recommend using them in async context.

(see also all other occurrences)

src/model/cloud-runner/remote-client/caching.test.ts
```diff
@@ -0,0 +1,65 @@
...
+      await Caching.PullFromCache(
+        cacheFolder.replace(/\\/g, `/`),
+        testFolder.replace(/\\/g, `/`),
+        `${CLI.options.cacheKey}`,
+      );
+      await CloudRunnerSystem.Run(`du -h ${__dirname}`);
+      await CloudRunnerSystem.Run(`tree ${testFolder}`);
+      await CloudRunnerSystem.Run(`tree ${cacheFolder}`);
+      // compare validity to original hash
+      expect(fs.readFileSync(path.resolve(testFolder, 'test.txt'), { encoding: 'utf8' }).toString()).toContain(
```
webbertakken, line 55: Please always put a new line above comments and return statements, unless it's the first thing in a method.

It improves both readability and git diffability by quite a bit.

src/model/cloud-runner/remote-client/setup-cloud-runner-repository.ts
```diff
@@ -0,0 +1,95 @@
...
+import path from 'path';
+import { assert } from 'console';
+import CloudRunnerLogger from '../services/cloud-runner-logger';
+import { CliFunction } from '../../cli/cli-functions-repository';
+import { CloudRunnerSystem } from '../services/cloud-runner-system';
+
+export class SetupCloudRunnerRepository {
+  public static async run() {
+    try {
+      await CloudRunnerSystem.Run(`mkdir -p ${CloudRunnerFolders.uniqueCloudRunnerJobFolderFull}`);
```
webbertakken, line 16: `run` is quite abstract and doesn't tell much about what's happening exactly.

In this case even though the class name tries being descriptive, it's still a bit hard to find out what the run method does exactly.

In this specific case you could consider calling a `CloudRunner` method `private loadRepository()` which could in turn use a `model class` or `service` called `Repository` (part of remote-client) which could expose a descriptive method of what it's doing exactly (or a high level method that calls descriptive private methods).

src/model/cloud-runner/services/cloud-runner-system.ts
```diff
@@ -0,0 +1,44 @@
+import { exec } from 'child_process';
+import { RemoteClientLogger } from '../remote-client/remote-client-logger';
+export class CloudRunnerSystem {
```
webbertakken, line 3: Please use an empty line after imports.

I'll introduce an ESLint rule that enforces this if it isn't already.

src/model/cloud-runner/services/cloud-runner-folders.ts
```diff
@@ -0,0 +1,73 @@
...
+    return path.join(CloudRunnerFolders.projectPathFull, `Library`);
+  }
+
+  public static get projectBuildFolderFull(): string {
+    return path.join(CloudRunnerFolders.repoPathFull, CloudRunner.buildParameters.buildPath);
+  }
+
+  public static get lfsDirectoryFull(): string {
+    return path.join(CloudRunnerFolders.repoPathFull, `.git`, `lfs`);
+  }
```
webbertakken, line 44: The way I see it a path is either relative or absolute, but always full.

Do we have any non-full examples? Would they be just the folderName?

src/model/cloud-runner/workflows/build-automation-workflow.ts
```diff
@@ -65,4 +60,62 @@ export class BuildAutomationWorkflow implements WorkflowInterface {
       throw error;
     }
   }
+
+  private static get FullWorkflow() {
```
webbertakken, line 64: perhaps `buildWorkflow`?

src/model/cloud-runner/workflows/build-automation-workflow.ts
```diff
@@ -65,4 +60,62 @@ export class BuildAutomationWorkflow implements WorkflowInterface {
...
+      ${setupHooks.filter((x) => x.hook.includes(`before`)).map((x) => x.commands) || ' '}
+      export GITHUB_WORKSPACE="${CloudRunnerFolders.repoPathFull.replace(/\\/g, `/`)}"
+      ${BuildAutomationWorkflow.SetupCommands(builderPath)}
+      ${setupHooks.filter((x) => x.hook.includes(`after`)).map((x) => x.commands) || ' '}
+      ${buildHooks.filter((x) => x.hook.includes(`before`)).map((x) => x.commands) || ' '}
+      ${BuildAutomationWorkflow.BuildCommands(builderPath, CloudRunner.buildParameters.buildGuid)}
+      ${buildHooks.filter((x) => x.hook.includes(`after`)).map((x) => x.commands) || ' '}`;
+  }
+
+  private static SetupCommands(builderPath) {
```
webbertakken, line 85: Note that in JavaScript and TypeScript the convention is to make all methods, variables, constants etc. `camelCase`.

We never use PascalCase except when using a `model class` or perhaps some kind of unit describing an `entity` or it's data directly.

src/model/input.ts
```diff
@@ -14,64 +14,66 @@ const core = require('@actions/core');
...
+    }
+    if (CloudRunnerQueryOverride.query(query, alternativeQuery)) {
+      return CloudRunnerQueryOverride.query(query, alternativeQuery);
+    }
+    if (process.env[query] !== undefined) {
+      return process.env[query];
+    }
+    if (alternativeQuery !== query && process.env[alternativeQuery] !== undefined) {
+      return process.env[alternativeQuery];
     }
```
webbertakken, line 40: Please use empty lines between if statements unless they're one-liners. Optionally with a comment above them (or ideally they'd be self-descriptive because they'd call methods from the domain model classes.

webbertakken (approved):
Amazing work once again. A huge step forward!

src/model/cli/cli.ts
frostebite: Agreed. 

When I started this work it was not clear the extent of the complexity of the configuration domain would be needed in total.

This is a fully dynamic / generic implementation.

We can harden to a specific implementation somewhat. But it does relate to a super set of input. 

So it would  potentially benefit from alterations to input as well, as in encapsulating it within a wider configuration concept, and this also made me want to hold off to start with.

src/model/cloud-runner/remote-client/setup-cloud-runner-repository.ts
frostebite: Delicious

src/model/cloud-runner/services/cloud-runner-folders.ts
frostebite: Agreed, it should be absolute or relative to be clear where needed. 

Uses of full will be removed.

src/model/cloud-runner/workflows/build-automation-workflow.ts
frostebite: Whatever I have been doing most that day comes with me sometimes. 😁🙂

webbertakken: Haha completely understand of course!

I'll set up some ESLint rules to mitigate this problem.

src/model/cli/cli.ts
webbertakken: Whatever you think is best here. We can always iterate as well.

frostebite: Created an issue to follow up on this and address it thoroughly

src/model/cloud-runner/cloud-runner.test.ts
```diff
@@ -4,7 +4,17 @@ import Input from '../input';
...
+import CloudRunnerLogger from './services/cloud-runner-logger';
 
+function guid() {
+  return Math.trunc((1 + Math.random()) * 0x10000)
+    .toString(16)
+    .slice(1);
+}
+function guidGenerator() {
+  return `${guid() + guid()}-${guid()}-${guid()}-${guid()}-${guid()}${guid()}${guid()}`;
+}
```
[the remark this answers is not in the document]
frostebite: as in https://www.npmjs.com/package/uuid specifically?

Yes it would, this was an alternative to another depedency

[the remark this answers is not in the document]
webbertakken: Following [this answer](https://stackoverflow.com/a/2117523/3593896) on stack overflow I'd go with that package indeed. As it's well tested and well supported.

src/model/cloud-runner/remote-client/setup-cloud-runner-repository.ts
frostebite: Renamed to `RemoteClient.bootstrapRepository`.

Within `remote-client/index.ts`

PR #355, opened 2022-03-11
feat: compatibility with self-hosted runners with SELinux

paulpach:
When using a self-hosted runner with SELinux (fedora)
volumes need to be mounted with ":z" in order to have write access
these flags are documented [here](https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label)

Also the in self-hosted runners, _temp is empty, so this creates the child folders.

#### Changes

- mount volumes with SELinux labels
- create folders _github_home and _github_workflow if they are missing

#### Checklist

### DIFF
--- a/src/model/docker.ts
+++ b/src/model/docker.ts
@@ -45,11 +45,9 @@ class Docker {
     switch (baseOs) {
       case 'linux':
         const github_home = join(runnerTemporaryPath, "_github_home");
-        if (!existsSync(github_home))
-          mkdirSync(github_home);
+        if (!existsSync(github_home)) mkdirSync(github_home);
         const github_workflow = join(runnerTemporaryPath, "_github_workflow");
-        if (!existsSync(github_workflow))
-          mkdirSync(github_workflow);
+        if (!existsSync(github_workflow)) mkdirSync(github_workflow);
 
         return `--env UNITY_SERIAL \
                 --env GITHUB_WORKSPACE=/github/workspace \
--- a/src/model/docker.ts
+++ b/src/model/docker.ts
@@ -1,6 +1,8 @@
 import { exec } from '@actions/exec';
 import ImageTag from './image-tag';
 import ImageEnvironmentFactory from './image-environment-factory';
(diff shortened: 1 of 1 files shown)


src/model/docker.ts
```diff
@@ -42,13 +44,18 @@ class Docker {
   static getBaseOsSpecificArguments(baseOs, workspace, unitySerial, runnerTemporaryPath, sshAgent): string {
     switch (baseOs) {
       case 'linux':
+        const github_home = join(runnerTemporaryPath, "_github_home");
+        existsSync(github_home) || mkdirSync(github_home);
+        const github_workflow = join(runnerTemporaryPath, "_github_workflow");
+        existsSync(github_workflow) || mkdirSync(github_workflow);
```
webbertakken, line 50: I believe these directories should exist according to the spec. Could you explain a little bit about why this is needed?

Also prefer the more semantic `if(condition) // do something` over `||` which makes the code harder to read for new developers

src/model/docker.ts
```diff
@@ -42,13 +44,20 @@ class Docker {
   static getBaseOsSpecificArguments(baseOs, workspace, unitySerial, runnerTemporaryPath, sshAgent): string {
     switch (baseOs) {
       case 'linux':
+        const github_home = join(runnerTemporaryPath, "_github_home");
+        if (!existsSync(github_home))
+          mkdirSync(github_home);
+        const github_workflow = join(runnerTemporaryPath, "_github_workflow");
+        if (!existsSync(github_workflow))
+          mkdirSync(github_workflow);
```
webbertakken, line 52:
```suggestion
        if (!existsSync(github_home)) mkdirSync(github_home);
        const github_workflow = join(runnerTemporaryPath, "_github_workflow");
        if (!existsSync(github_workflow)) mkdirSync(github_workflow);
```

webbertakken:
Could you run `yarn husky install`? Looks like none of the hooks are working for you.

webbertakken:
Thank you very much!

PR #357, opened 2022-03-13
Attempt to streamline dev lifecycle (git hooks)

webbertakken:
#### Changes

- Introduce [Lefthook](https://github.com/evilmartians/lefthook) ([reference](https://github.com/evilmartians/lefthook/blob/master/docs/full_guide.md))
- Remove Husky and Lint-staged (reduce the number of dependencies by 30 dependencies)
- Update hooks to not do unnecessary work
- make them executable
- Ignore lib; jest was taking mocks from there.
- Ignore dist in hooks (was taking some time to process large dist files)
- Remove extreme checks from pre-commit hook 
  - only related tests will be run) 
  - only changed files checked against code styles.
- Make cross-platform compatible
- Fail when something gets logged to console in tests (use `jest.spyOn` instead)
- Correct ~any~ some tests that assume some tools are installed on local machines

#### Checklist

- [x] Check that this works on Windows (Powershell 7 or WSL)
- [ ] Check that this works on MacOS
- [ ] Check that this works on Linux


.husky/pre-commit
```diff
@@ -1,7 +1,6 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-yarn lint-staged
-yarn lint
-yarn test
+yarn lint-staged --verbose
 yarn build
+git add dist
```
davidmfinol, line 6: Doesn't this cause warnings?

webbertakken: Yea it needs some more work. I spent a lot of time making it work smoothly across all platforms before we upgraded.

I'm a bit disappointed that I (and others) weren't able to easily contribute anymore. Having to also deal with dev life cycle stuff makes contributing very hard.

webbertakken: replaced `husky` and `lint-staged` altogether, using `lefthook` instead.

davidmfinol (approved):

PR #362, opened 2022-03-26
feat: work with rootless docker

paulpach:
Running docker currently mounts the docker.sock file into the container.
This was introduced in 2ab738c08389696888fa13b09fb29d02932a00b3 but
there is no explanation provided.

The docker.sock file is only needed if we want to run docker inside the container
to create other images or start other containers.
I searched through the code and I did not find any such use.

In particular, on fedora this gives permission denied because docker.sock
is owned by root and the container runs under an unprivileged user.
One has to change the permissions of docker.sock
(which is actually a link to /run/podman/podman.sock) to be writeable by the user.

If we don't need to use docker inside the containers,  then we can remove this file,
thus we can run this GitHub action as an unprivileged user out of the box.

#### Changes

- Remove mounting docker.sock in the container

#### Checklist

### DIFF
--- a/src/model/docker.ts
+++ b/src/model/docker.ts
@@ -52,7 +52,6 @@ class Docker {
         return `--env UNITY_SERIAL \
                 --env GITHUB_WORKSPACE=/github/workspace \
                 ${sshAgent ? '--env SSH_AUTH_SOCK=/ssh-agent' : ''} \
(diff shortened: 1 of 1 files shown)


paulpach:
I just checked, docker is not even installed inside those images,  so having docker.sock is not useful at all.

webbertakken (approved):
Thanks for the clear description. Looks good to me!

PR #365, opened 2022-03-30
perf: avoid building docker image

paulpach:
mount the scripts using volume instead of building docker image with the scripts.

in my setup,  (self hosted runner on fedora) the build step goes from 6:23 to 2:04

#### Checklist

- [ ] Windows builds


src/model/docker.ts
```diff
@@ -36,12 +37,13 @@ class Docker {
     --rm \
     ${ImageEnvironmentFactory.getEnvVarString(parameters)} \
     ${baseOsSpecificArguments} \
-    ${image}`;
+    ${image} -- \
```
webbertakken, line 40: I would keep this on two lines, or put the two hyphens on the line that passed the actual command
```suggestion
    ${image} \
    -- \
```

src/index.ts
```diff
@@ -27,8 +27,8 @@ async function runMain() {
       if (process.platform === 'darwin') {
         MacBuilder.run(actionFolder, workspace, buildParameters);
       } else {
-        builtImage = await Docker.build({ path: actionFolder, dockerfile, baseImage });
-        await Docker.run(builtImage, { workspace, ...buildParameters });
+        //builtImage = await Docker.build({ path: actionFolder, dockerfile, baseImage });
+        await Docker.run(baseImage, { workspace, actionFolder, ...buildParameters });
```
webbertakken, line 31: We have to either split between platforms or make Windows compatible with the same changes here.

webbertakken (changes requested):
Excellent performance fix!

src/model/docker.ts
webbertakken: (already committed)

src/index.ts
paulpach: yeap,  I haven't touched windows yet,  I wanted feedback from you to see if this approach was good or not.

webbertakken (approved):

PR #371, opened 2022-03-31
feat: set activate command to always return exit code 0

EricRibeiro:
This change will allow pipelines with shells that exit on an error to run the activate script.

#### Changes

- Set the activate command to always return true.
- Replace log output from `/dev/stdout` with a file.
-  Switch `echo` with `cat` to print the command output.

#### Checklist

### DIFF
--- a/dist/platforms/ubuntu/steps/activate.sh
+++ b/dist/platforms/ubuntu/steps/activate.sh
@@ -28,10 +28,14 @@ if [[ -n "$UNITY_LICENSE" ]] || [[ -n "$UNITY_LICENSE_FILE" ]]; then
   fi
 
   # Activate license
-  ACTIVATION_OUTPUT=$(unity-editor \
-      -logFile /dev/stdout \
-      -quit \
-      -manualLicenseFile $FILE_PATH)
+  # Set to always return true, otherwise the shell will exit on pipelines set to fail on error
+  ACTIVATION_OUTPUT=UnityActivation.log
+
+  unity-editor \
+    -logFile "$ACTIVATION_OUTPUT" \
+    -quit \
+    -manualLicenseFile $FILE_PATH \
+    || true
 
   # Store the exit code from the verify command
   UNITY_EXIT_CODE=$?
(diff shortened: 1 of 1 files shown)


dist/platforms/ubuntu/steps/activate.sh
```diff
@@ -28,10 +28,14 @@ if [[ -n "$UNITY_LICENSE" ]] || [[ -n "$UNITY_LICENSE_FILE" ]]; then
...
-      -quit \
-      -manualLicenseFile $FILE_PATH)
+  # Set to always return true, otherwise the shell will exit on pipelines set to fail on error
+  ACTIVATION_OUTPUT=UnityActivation.log
+
+  unity-editor \
+    -logFile "$ACTIVATION_OUTPUT" \
+    -quit \
+    -manualLicenseFile $FILE_PATH \
+    || true
```
webbertakken, line 38: Note that unity-builder is made specifically for running on GitHub Actions but I'm assuming you know that.

That would work, but personally I wouldn't pollute our code with the assumption that the editor is always going to return exit code 1 (bug in unity). Perhaps in some versions the exit code starts working.

Therefor I believe a comment like on line 31 is better in case of `unity-builder` specifically.

dist/platforms/ubuntu/steps/activate.sh
```diff
@@ -43,7 +47,7 @@ if [[ -n "$UNITY_LICENSE" ]] || [[ -n "$UNITY_LICENSE_FILE" ]]; then
   #
   #   "LICENSE SYSTEM [2020120 18:51:20] Next license update check is after 2019-11-25T18:23:38"
   #
-  ACTIVATION_SUCCESSFUL=$(echo $ACTIVATION_OUTPUT | grep 'Next license update check is after' | wc -l)
+  ACTIVATION_SUCCESSFUL=$(cat $ACTIVATION_OUTPUT | grep 'Next license update check is after' | wc -l)
```
webbertakken, line 50: Ah nice find. Did this break when we introduced an actual logfile?

webbertakken (changes requested):

PR #383, opened 2022-04-11
Fix iOS Fastlane build when using Unity Cloud Diagnostics

gnomesley:
Fixes error described here: https://forum.unity.com/threads/ios-build-is-failing-seems-like-a-fastlane-problem-not-sure-how-to-proceed.682201/

#### Changes

- Add username and password to mac build.sh

#### Checklist

### DIFF
--- a/dist/platforms/mac/steps/build.sh
+++ b/dist/platforms/mac/steps/build.sh
@@ -130,6 +130,8 @@ echo ""
   -quit \
   -batchmode \
   -nographics \
+  -username "$UNITY_EMAIL" \
+  -password "$UNITY_PASSWORD" \
   -customBuildName "$BUILD_NAME" \
   -projectPath "$UNITY_PROJECT_PATH" \
   -buildTarget "$BUILD_TARGET" \


dist/platforms/mac/steps/build.sh
```diff
@@ -130,6 +130,8 @@ echo ""
   -quit \
   -batchmode \
   -nographics \
+  -username "$UNITY_EMAIL" \
+  -password "$UNITY_PASSWORD" \
```
webbertakken, line 134: These are already passed during the activate step. Any reason why we'd need to pass them here too?

webbertakken:
The description is not clear enough as to what exactly it fixes under which circumstances.

We currently have successful builds in iOS.

What would be interesting to know is how `Unity Cloud Diagnostics` is different and which exact post you're referring to in that long thread that is seemingly about a bug in Unity and little related to the code changes. I could be mistaken and would love to be educated.

gnomesley:
The post I'm referring to is the first post in the thread.
I get this error during the `build_app` step in Fastlane:
```
6842: ▸ Running script 'Process symbols'
6843: ▸ ** ARCHIVE FAILED **
6844: ▸ The following build commands failed:
6845: ▸ PhaseScriptExecution Process\ symbols /BUILD_PATH/Library/Developer/Xcode/DerivedData/Unity-iPhone-gjomesfkojewdbcqkxboenqtbzba/Build/Intermediates.noindex/ArchiveIntermediates/Unity-iPhone/IntermediateBuildFilesPath/Unity-iPhone.build/Release-iphoneos/Unity-iPhone.build/Script-F6E34F12805380276D47C8E3.sh
6846: ▸ (1 failure)
```

The [solution](https://forum.unity.com/threads/ios-build-is-failing-seems-like-a-fastlane-problem-not-sure-how-to-proceed.682201/#post-4728452) from the thread is to add the username and password arguments when calling Unity through batchmode.

webbertakken (approved):
Thanks for the additional context.

Since for MacOS the `UNITY_EMAIL` and `UNITY_PASSWORD` are always passed we don't need conditionals for whether or not to pass them to the CLI.

Therefor I think we're safe to merge this.

gnomesley:
Sure thing, I'll let you know when the build is done.

gnomesley:
Looks like it worked. 🙂

PR #387, opened 2022-04-16
Cloud runner v0.2 - continued quality of life improvements

frostebite:
#### Changes

Cloud runner latest tweaks. 
Most noteworthy being 
- support for an exit early flag that skips cleanup, allowing you run ephemeral self-hosted runners for github in a cloud runner job. You can then run your main workflow and skip the majority of the runner minutes.
- garbage collection options for aws cli
- include webgl in build pipeline for cloud runner
- cache uses tar instead of zip for speed - plan to introduce optional compression
- only run input overrides for cloud runner

#### Checklist


src/integrity.test.ts
```diff
@@ -1,9 +1,9 @@
-import { stat } from 'fs/promises';
```
webbertakken, line 1: I'd say the previous import style is preferable, because it (normally) allows for treeshaking

```
import { namedImport } from 'fs/promises';
```

frostebite: It was failing the build for some reason. Will change it back and dig into why.

src/model/input-readers/git-repo.ts
```diff
@@ -14,6 +18,9 @@ export class GitRepoReader {
   }
 
   public static async GetBranch() {
+    if (Input.cloudRunnerCluster === 'local') {
+      return '';
+    }
```
webbertakken, line 23: As a general note, if you have safeguard conditions that immediately branch off with a return you can also make it into a one-liner and leave an empty line after. I usually do this for very simple statements to further improve readability.

webbertakken (approved):
Looking great!

src/integrity.test.ts
webbertakken: I also started using `fs/promises` quite recently so it could be that the linter complains because I haven't updated the rules yet.

frostebite: image

frostebite: @webbertakken any idea what the issue is?

webbertakken: Yup, you need to `nvm use 16.13.0` or similar to use a newer node version.

Some of our repos still require 12, but we should upgrade to latest stable.

PR #395, opened 2022-05-01
Cloud runner develop v0.1

frostebite:
- Cancel follow logs flag now requires `cloudRunnerIntegrationTests` flag to avoid accidental triggering
- Link to AWS Cloud Watch logs fixed
- Cleanup pipeline now uses AWS list all CLI command which includes info about resource age for monitoring


frostebite:
Currently the AWS pipelines fail because I deleted the base stack to cleanup and test from scratch.

But this doesn't delete the AWS VPCs it seems, now I have reached the max 5 I can create. And deleting them is blocked because you must first delete dependent resources.

I will work through that shortly and reinitialize the base stack.

davidmfinol (approved):

PR #418, opened 2022-07-28
WIP: macOS module support & Android Library Support

Legoless:
#### Changes

This is a WIP PR (as we needed the fixes for our processes), as I didn't have enough time to go through the checklist and Windows is still missing Android Library support.

So there are several changes:
- Injected build script now uses `exportAsGoogleAndroidProject` setting and applies it correctly.
- Injected build script now applies `buildAppBundle ` setting, if set specifically. Default remains that it's read from `.aab` in build path filename. This enables the setting to be applied while exporting the project.
-  If `exportAsGoogleAndroidProject` is enabled, filename doesn't include `.aab` or `.apk` anymore, thus giving us backwards support, otherwise the behaviour is the same as previously. Enable the setting in your workflow: `exportAsGoogleAndroidProject: true`
- MacOS Setup now correctly installs modules for Mac, iOS, tvOS and Android, which allows you to perform all of those builds (for some reason, Ubuntu on GitHub Actions has issues building iOS projects, but Mac works, didn't look into details, but scenes are corrupted and game crashes at launch time).

Let me know your feedback and proposed changes. I'll also need help with testing, as we don't use Ubuntu machines much. Use `blubblub/unity-builder@2.2.0` to test the action.

#### Checklist


src/model/image-environment-factory.ts
```diff
@@ -48,8 +48,8 @@ class ImageEnvironmentFactory {
       { name: 'ANDROID_KEYALIAS_PASS', value: parameters.androidKeyaliasPass },
       { name: 'ANDROID_TARGET_SDK_VERSION', value: parameters.androidTargetSdkVersion },
       { name: 'ANDROID_SDK_MANAGER_PARAMETERS', value: parameters.androidSdkManagerParameters },
-      { name: 'ANDROID_APP_BUNDLE', value: parameters.androidAppBundle },
-      { name: 'EXPORT_AS_GOOGLE_ANDROID_PROJECT', value: parameters.exportAsGoogleAndroidProject },
+      { name: 'ANDROID_APP_BUNDLE', value: parameters.androidAppBundle ? 'true' : 'false' },
+      { name: 'EXPORT_AS_GOOGLE_ANDROID_PROJECT', value: parameters.exportAsGoogleAndroidProject ? 'true' : 'false' },
```
webbertakken, line 52: Values for these parameters should be determined in `Input` or `Parameters`, not directly in the file that passes them to docker.

Legoless: Not sure what you mean, both are pulled from `parameters` and both are only boolean values, so it looked most clear to me that, these are sent into environment as boolean strings.

webbertakken: the image-environment-factory isn't meant to manipulate the values of the parameters. That's what the other two files are for (where the parameters come from).

src/model/platform-setup/setup-mac.ts
```diff
@@ -30,13 +30,41 @@ class SetupMac {
...
+      case 'iOS':
+        command += '--module ios ';
+        break;
+      case 'tvOS':
+        command += '--module tvos ';
+        break;
+      case 'Android':
+        command += '--module android ';
+        break;
+      case 'MacOS':
```
johnfoconnor, line 45: this case needs to support `StandaloneOSX` MacOS isnt a valid targetPlatform for unity. see values here https://docs.unity3d.com/ScriptReference/BuildTarget.html

src/model/platform-setup/setup-mac.ts
```diff
@@ -30,13 +30,41 @@ class SetupMac {
...
+        command += '--module mac-il2cpp ';
+        break;
+      case 'WebGL':
+        command += '--module webgl ';
+        break;
+      default:
+        throw new Error(`Unsupported module for target platform: ${targetPlatform}.`);
```
johnfoconnor, line 52: I think it's reasonable to add an explicit `unityModules` input which is a list of N modules and would be additive to the output of this case statement. and we only raise exception if the modules list empty.

e.g. as of 2020.2 some modules exist for for localization (e.g. `language-ja`, `language-ko`). These are complementary and completely unrelated to the targetPlatform

https://forum.unity.com/threads/package-ui-localization-is-available-in-2020-2.957173/

webbertakken:
Is this something you intend to finish or shall we close?

PR #303, opened 2021-12-10
Support version tags that don't start with v for semantic versioning

davidmfinol:
#### Changes

Fixes #242

#### Checklist

- [x] Read the contribution [guide](../CONTRIBUTING.md) and accept the [code](../CODE_OF_CONDUCT.md) of conduct
- [x] Readme (updated or not needed)
- [x] Tests (added, updated or not needed)

PR #420, opened 2022-08-02
Remove v prefix from tag checking

abbydiode:
### Description

There is no mention of requiring a prefix to the tags in the [documentation](https://game.ci/docs/github/builder#versioning). This is unexpected behavior, and as a result of this requirement tags such as `0.1.0` do not work with the semantic versioning strategy. Additionally, the [semantic versioning](https://semver.org) documentation referenced has not used the prefix for a couple of years.

### Changes

- Removed the requirement to have a `v` prefixing tags.

### DIFF
--- a/dist/default-build-script/Assets/Editor/UnityBuilderAction/Versioning/Git.cs
+++ b/dist/default-build-script/Assets/Editor/UnityBuilderAction/Versioning/Git.cs
@@ -40,7 +40,7 @@ namespace UnityBuilderAction.Versioning
     /// </summary>
     public static string GetTagVersion()
     {
-      string version = Run(@"tag --points-at HEAD | grep ?[0-9]*");
+      string version = Run(@"tag --points-at HEAD | grep *[0-9]*");
 
       version = version.Substring(1);
 
@@ -62,7 +62,7 @@ namespace UnityBuilderAction.Versioning
     /// </summary>
     static bool HasAnyVersionTags()
     {
-      return "0" != Run(@"tag --list --merged HEAD | grep ?[0-9]* | wc -l");
+      return "0" != Run(@"tag --list --merged HEAD | grep *[0-9]* | wc -l");
     }
 
     /// <summary>
@@ -92,7 +92,7 @@ namespace UnityBuilderAction.Versioning
     /// </summary>
     static string GetVersionString()
     {
-      return Run(@"describe --tags --long --match ""?[0-9]*""");
+      return Run(@"describe --tags --long --match ""*[0-9]*""");
 
       // Todo - implement split function based on this more complete query
       // return Run(@"describe --long --tags --dirty --always");
--- a/dist/default-build-script/Assets/Editor/UnityBuilderAction/Versioning/Git.cs
+++ b/dist/default-build-script/Assets/Editor/UnityBuilderAction/Versioning/Git.cs
@@ -40,7 +40,7 @@ namespace UnityBuilderAction.Versioning
     /// </summary>
     public static string GetTagVersion()
     {
-      string version = Run(@"tag --points-at HEAD | grep [0-9]*");
+      string version = Run(@"tag --points-at HEAD | grep ?[0-9]*");
 
       version = version.Substring(1);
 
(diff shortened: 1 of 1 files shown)


webbertakken:
Hi @abbydiode, thank you for raising a PR to improve things!

Agreed that this is somewhat confusing. 

However we determined some time ago that it is convention for version tags on GitHub start with a `v`. 
Since this has worked well for a few years now and because this would introduce a breaking change I believe it's better to update the documentation instead of the behaviour itself.

abbydiode:
> However we determined some time ago that it is convention for version tags on GitHub start with a `v`. Since this has worked well for a few years now and because this would introduce a breaking change I believe it's better to update the documentation instead of the behaviour itself.

@webbertakken thank you for your feedback. I thought the way the changes were made the `v` prefix would be optional. However I realized afterwards that I had made a mistake with the matching pattern. This has been resolved in the current state of this PR, meaning the `v` prefix is now optional.

webbertakken:
I really appreciate the effort, but as mentioned it might be better to fix the documentation instead of the behaviour because it worked well (other than the confusion you mentioned) and the change would introduce a breaking change: anyone who upgrades to the new version will now have to remove all non `v` version tags that start with a number if they weren't intended for versioning.

Closing for that reason.

connieprice:
> I really appreciate the effort, but as mentioned it might be better to fix the documentation instead of the behaviour because it worked well (other than the confusion you mentioned) and the change would introduce a breaking change: anyone who upgrades to the new version will now have to remove all non `v` version tags that start with a number if they weren't intended for versioning.
> 
> Closing for that reason.

I don't see how this is a breaking change, or where including v is a GitHub standard. Making 'v' optional should not break any existing workflows, and would allow for new workflows to follow the semantic versioning standard more closely.

abbydiode:
I believe that if a users assigns for example `1.2.3` as a tag to a commit it would be expected behavior for the builder's semantic versioning to pick up on that. Not only if they assign for example `v1.2.3`.

Additionally I would like to add that the [FAQ](https://semver.org/#faq) for semantic versioning specifically mentions that anything including `v` is not a semantic version:

> **Is “v1.2.3” a semantic version?**
No, “v1.2.3” is not a semantic version. However, prefixing a semantic version with a “v” is a common way (in English) to indicate it is a version number. Abbreviating “version” as “v” is often seen with version control. Example: git tag v1.2.3 -m "Release version 1.2.3", in which case “v1.2.3” is a tag name and the semantic version is “1.2.3”.

webbertakken:
I understand where you're coming from. We've had this same discussion a few times before on [our Discord](https://game.ci/discord).

**Semantic version vs version tags**
Don't get me wrong, I'm not saying `v1.2.3` is a semantic version. I'm saying it's most commonly used here on github to denote the tag holds a version number. I also didn't invent that, but years ago we decided to follow that best practice with unity-builder.

**Why is it a breaking change?**
Please note that this action is used by many projects. Suddenly counting other commits to count for versioning may mess with people's existing workflows. Therefore we'd have to mark it as a breaking change and bump major for it. Note that asking people to remove existing tags might also be problematic.

webbertakken:
Just looking through history and it looks like we already voted in favour of making the `v` optional afterall.

https://github.com/game-ci/unity-builder/pull/303

What you're looking for already works as expected.

abbydiode:
> Please note that this action is used by many projects. Suddenly counting other commits to count for versioning may mess with people's existing workflows. Therefore we'd have to mark it as a breaking change and bump major for it. Note that asking people to remove existing tags might also be problematic.

Sorry, I'm not sure if I understand why users would have to count 'other commits' for versioning and why users would have to remove existing tags. All this change would introduce is the `v` prefix being optional or not.

To me the fact that only tags with the `v` prefix are counted to increase the major/minor version with the semantic versioning strategy is problematic for the following reasons:
1. It is not obvious from the documentation that the `v` prefix is necessary, therefore it may look to users like this versioning strategy is broken.
2. When requiring the `v` prefix your list of tags will get cluttered with manually assigned major/minor increase tags that have the `v` prefix. Whereas the output from the build step does not have the `v` prefix. To rectify this in a CI pipeline users will always have to include the `v` in their actions.

Even if this would be considered a breaking change, wouldn't it be possible to create a 'new' versioning strategy that's essentially the same as the regular semantic versioning strategy but with the `v` prefix removed entirely to more closely follow the semantic versioning standard?

webbertakken:
> Sorry, I'm not sure if I understand why users would have to count 'other commits' for versioning and why users would have to remove existing tags. All this change would introduce is the v prefix being optional or not.

Sorry that was a typo, I meant to include more tags in versioning. with `other commits` I meant `additional tags`. Edited.


About the rest of your arguments, see my last post. This already works exactly how you describe you expect it to work.

PR #423, opened 2022-08-02
fix: make v character in version tags optional

webbertakken:
#### Changes

- Fixes a bug that did not allow tags without `v` to be recognised using the automatic versioning strategy (`versioning: Semantic`)
- Original change: https://github.com/game-ci/unity-builder/pull/303
- Addresses https://github.com/game-ci/unity-builder/pull/420

#### Checklist


abbydiode:
I have tested this and while it did correctly identify the existing tag, it did not auto-increment the patch version.

webbertakken:
> I have tested this and while it did correctly identify the existing tag, it did not auto-increment the patch version.

It looks like it's actually using your `0.1.0` tag

image

Please try to create a tag that's higher, like `v0.2` or `0.2` to rule out flakiness.

abbydiode:
> Please try to create a tag that's higher, like v0.2 or 0.2 to rule out flakiness.

You're right, the previously generated 0.1.0 tag was messing with the versioning. However this does make me worry whether this would happen again if the versioning generated a new version with a patch number of 0.

abbydiode:
Nevermind! I've [verified](https://github.com/abbydiode/Accession/runs/7642833785) that versioning also works after generating a tag with patch number 0. I'm not sure why it worked for the `0.2.0` tag that it generated but not for the `0.1.0` tag that it previously generated. But it does work!

abbydiode:
Little update; I've worked out why it worked for `0.2.0` but not for `0.1.0`.

My manually assigned `0.1` tag was on a commit that was made earlier than the generated `0.1.0` tag. When I assigned `0.2` the `0.2.0` tag was also assigned to that same commit.

This may be an issue when the system generates a tag with patch number 0 that is not on the same commit as the commit with the manually assigned tag, however I am not sure if that is possible in the first place.

webbertakken:
We're working on a best effort basis here. Edge cases that may never even happen can always be covered in next interactions.

Based on your findings it looks like we should be good with this change.

I will add tests when I have time to make sure this doesn't accidentally break.

webbertakken:
Added tests for the regex. The tests are actually running grep to make sure it works using `grep -E`. These tests work on Windows (local) and Ubuntu (GitHub workflow). Accepted values as tags that are usable for our automated versioning strategy (when `versioning: Semantic` which is the default if `versioning` parameter is not set as well).

Accepted values are partial tags and any completely valid semantic version tag, as well as all variants prepended with a `v`.

image
image


Not accepting some weirder cases out of the incorrect semantic versions list, even if they are prepended with a `v`.

image

davidmfinol (approved):

PR #468, opened 2022-10-17
Feature/support for unity licensing server linux

simensan:
Hi,

I needed support for a lot of parallel builds and as such also wanted to use the build server as it unlocks using cheaper licenses that can only be used by CI ([https://unity.com/products/unity-build-server](https://unity.com/products/unity-build-server)). Since game-ci did not currently support this, I have forked and implemented the changes as bare-bones as possible to the level where it currently works for us.  Since this might be useful to others I hope to be able to contribute this back to gameci. I am no bash-script-guru, and this was my first time working with TS. So I probably haven't solved it all gracefully, but I am happy to make any fixes/improvements you come up with. I have also only fixed this for the ubuntu platform, but looking at the other platforms they already do not seem to support all features, like using license files. I am happy to try to add this to mac also if desired, but I assume that almost all unity builds will be done on the ubuntu platform to save build-minutes, so I have focused on that initially. 

I have added a template for the necessary services-config.json file here: dist/resources/services-config.json.template
I am not sure if there is a better location for it? Alternatively it can just be inlined in some script. 

#### Changes
- Added UNITY_LICENSING_SERVER build parameter. 
- Added tests to src/model/build-parameters.test.ts for new parameter
- Clear mocks between tests in src/model/build-parameters.test.ts 
- Fixed cloud runner tests to work with remote repo being a ssh url
- activate.sh will request a floating license, and return_license.sh will return it. 

#### Checklist
- [x] Readme - not needed
- [x] Tests - added where I found applicable


dist/platforms/ubuntu/steps/activate.sh
```diff
@@ -74,6 +74,26 @@ elif [[ -n "$UNITY_SERIAL" && -n "$UNITY_EMAIL" && -n "$UNITY_PASSWORD" ]]; then
...
+  pushd /resources
+  cat services-config.json.template | tr -d '\r' | awk "{sub(/%URL%/,\"$UNITY_LICENSING_SERVER\")}1" > services-config.json
+  mkdir -p /usr/share/unity3d/config/
+  mv services-config.json /usr/share/unity3d/config/
+
+  /opt/unity/Editor/Data/Resources/Licensing/Client/Unity.Licensing.Client --acquire-floating > license.txt #is this accessible in a env variable?
+  PARSEDFILE=$(grep -oP '\".*?\"' < license.txt | tr -d '"')
+  export FLOATING_LICENSE
+  FLOATING_LICENSE=$(sed -n 2p <<< "$PARSEDFILE")
+  FLOATING_LICENSE_TIMEOUT=$(sed -n 4p <<< "$PARSEDFILE")
```
webbertakken, line 91: The parsing and writing of this file should happen in the logic (javascript) part, not in bash. We're trying to keep the bash part as thin as possible because it's extremely hard to test and maintain.

webbertakken (changes requested):
Amazing work!

1 comment. The rest looks great!

dist/platforms/ubuntu/steps/activate.sh
simensan: I assume you are talking about the services-config.json? I can definitely fix that, but is there any way for me to copy it to the right location inside the container without mounting and doing it from bash? Or should I just mount it to /usr/share/unity3d/config/ in the docker command since I am creating the directory from script anyways?

simensan:
I have moved the processing of services-config file to platform-setup, let me know if you think there is a better place for it.

webbertakken:
> I have moved the processing of services-config file to platform-setup, let me know if you think there is a better place for it.

Sure! I like that it's part of the logic now. That makes it more manageable for us.
Personally I would argue it's more of a feature setup than platform setup so I would have created a new model for it specifically for this feature. But this is also good for a first iteration.

webbertakken (approved):

simensan:
> > I have moved the processing of services-config file to platform-setup, let me know if you think there is a better place for it.
> 
> Sure! I like that it's part of the logic now. That makes it more manageable for us. Personally I would argue it's more of a feature setup than platform setup so I would have created a new model for it specifically for this feature. But this is also good for a first iteration.

I agree, makes more sense. I will make sure to do an iteration on that, will need to make this change to the test runner also, so will do a take on a feature-setup there first.

webbertakken:
You have to rebase on main, or merge main into your branch and run yarn build to generate the dist files again. It can't be merged as is.

simensan:
Fixed

webbertakken:
Thanks so much for your contribution! 🧡

PR #478, opened 2022-11-22
feat: Android updates for Windows and androidVersionCode output

AndrewKahr:
#### Changes

- Create android keystore on windows when available
- Output android version code

Needs https://github.com/game-ci/documentation/pull/349

#### Checklist

### DIFF
--- a/dist/platforms/windows/build.ps1
+++ b/dist/platforms/windows/build.ps1
@@ -67,6 +67,21 @@ else
 }
 
 #
+# Create Android Keystore if available
+#
+
+if ( "$Env:BUILD_TARGET" -eq "Android" -and -not ([string]::IsNullOrEmpty("$Env:ANDROID_KEYSTORE_NAME")) -and -not ([string]::IsNullOrEmpty("$Env:ANDROID_KEYSTORE_BASE64")) )
+{
+    Write-Output "Creating Android keystore."
+    $keystorePath = "$Env:GITHUB_WORKSPACE\$Env:ANDROID_KEYSTORE_NAME"
+    [System.IO.File]::WriteAllBytes($keystorePath, [System.Convert]::FromBase64String($Env:ANDROID_KEYSTORE_BASE64))
+    Write-Output "Created Android keystore."
+}
+else {
+    Write-Output "Not creating Android keystore."
+}
+
+#
 # Pre-build debug information
 #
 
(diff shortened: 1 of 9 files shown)


dist/platforms/windows/build.ps1
```diff
@@ -66,6 +66,21 @@ else
...
+if ( "$Env:BUILD_TARGET" -eq "Android" -and -not ([string]::IsNullOrEmpty("$Env:ANDROID_KEYSTORE_NAME")) -and -not ([string]::IsNullOrEmpty("$Env:ANDROID_KEYSTORE_BASE64")) )
+{
+    Write-Output "Creating Android keystore."
+    $keystorePath = "$Env:GITHUB_WORKSPACE\$Env:ANDROID_KEYSTORE_NAME"
+    [System.IO.File]::WriteAllBytes($keystorePath, [System.Convert]::FromBase64String($Env:ANDROID_KEYSTORE_BASE64))
+    Write-Output "Created Android keystore."
+}
+else {
+    Write-Output "Not creating Android keystore."
+}
```
webbertakken, line 82: This should probably move into the TypeScript logic. We don't want to keep adding complexity in the (power)shell scripts.

webbertakken (changes requested):

davidmfinol (commented):
LGTM, just needs to resolve merge conflicts by pulling the main branch and rebuilding.

dist/platforms/windows/build.ps1
AndrewKahr: Should this be removed from the Linux build script as well and be done purely in js?

davidmfinol: This matches how its currently done for the Linux build script.

I think it's better to first make the Windows and Linux build scripts equivalent, and then we can go back and change both at the same time.

webbertakken: Yes ultimately this is logic not passing variables to the editor. All logic is intended to be in the TS part.

In my opinion we shouldn't first make it equivalent, because that would just make it more work for me (not to mention impossible to test) when porting it over to https://github.com/game-ci/cli.

AndrewKahr: I'll look into moving it to ts. Hopefully its relatively straightforward being a base64 conversion and a write to a file

AndrewKahr: Implemented in latest push

davidmfinol (approved):

webbertakken (approved):
Great!

PR #506, opened 2023-02-14
Add default value for androidExportType

davidmfinol:
#### Changes

- Add default value for androidExportType

#### Checklist

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -53,6 +53,7 @@ inputs:
     description: '[Deprecated] Use androidExportType instead. Whether to build .aab instead of .apk'
   androidExportType:
     required: false
+    default: 'androidPackage'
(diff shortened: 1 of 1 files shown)


AndrewKahr:
@davidmfinol Note this will cause a breaking change. The code is configured to prefer androidExportType over exportAndroidAppBundle. If you set a default for this now, anyone who hasn't set the parameter explicitly and is still using the old parameter will have an androidPackage built. I believe the issue that covered this preferred the appBundle parameter to be deprecated and later removed in v3. At that stage it would be ideal to add the default value.

davidmfinol:
@AndrewKahr Thanks for the catch! I'll revert the change.

PR #510, opened 2023-02-14
Fix defaults

AndrewKahr:
#### Changes

- ...

#### Checklist

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -53,6 +53,7 @@ inputs:
     description: '[Deprecated] Use androidExportType instead. Whether to build .aab instead of .apk'
   androidExportType:
     required: false
+    default: ''
     description:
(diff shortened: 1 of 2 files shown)


davidmfinol:
Is there a reason to completely remove the `default` line for `androidExportType`, instead of doing `default: ''`?

AndrewKahr:
I'm not sure exactly how inputs works. I'm relying on it not being defined to know whether to prioritize over the app bundle parameter. If '' counts as undefined then it can be added as default.

AndrewKahr:
@davidmfinol Is it preferred to have an empty string default? I can see about changing the logic to support that.

davidmfinol:
I think it would be better to have to the empty string default, so that it is consistent with how we are defining all the other inputs.

davidmfinol (approved):

davidmfinol (approved):

PR #511, opened 2023-02-17
Code cleanup

AndrewKahr:
Functionally there should be no changes to end users

#### Changes

- Enable noImplicitAny
- Add types to all implicit any variables
- Bump target to ES2020 for recent language features (optional chaining)
- Code cleanup
- Add debug configuration for vscode
- Remove the autorun flag from jest to remove warning
- Bump packages to fix dependency version mismatch warning 
- Changed @arkweid/lefthook to @evilmartians/lefthook as @arkweid/lefthook has been deprecated in favor of @evilmartians/lefthook 
- Added concurrency groups to integrity check and build workflows. New commits to branches/PRs will cancel superseded runs on the same branch/PR
- Update imports to not use require syntax
- Use node packages (ie node:fs rather than fs)
- AndroidVersionCode is now a string rather than a number as it gets converted to a string when passed out of the system
- Reduce timeout for windows builds
- Remove 2020.1.17f1 from windows builds due to repeated license activation errors 
- Update the naming scheme of workflows for consistency
- Update build names so the target platform and unity version aren't cut off by GitHub actions UI

#### Checklist


AndrewKahr:
@frostebite (I would request a review but unfortunately that option isn't available, forgive the ping)
To reenable `noImplicitAny`, I had to go through cloudrunner to update the types, so I think it would be good if you took a look in addition to @davidmfinol to ensure I haven't broken anything. The unit tests pass so I would expect it to work but I have no way of verifying as I do not have access to any cloud clusters.

frostebite:
> @frostebite (I would request a review but unfortunately that option isn't available, forgive the ping)
> To reenable `noImplicitAny`, I had to go through cloudrunner to update the types, so I think it would be good if you took a look in addition to @davidmfinol to ensure I haven't broken anything. The unit tests pass so I would expect it to work but I have no way of verifying as I do not have access to any cloud clusters.

Sure

davidmfinol (changes requested):
Can we add 
```
    exclude:
      - targetPlatform: Android
        unityVersion: 2022.2.6f1
```
to `build-tests-ubuntu.yml`?
Just to have all the tests pass until we are able to have a fix for Android on 2022?

davidmfinol (approved):

davidmfinol:
@frostebite @webbertakken Do you guys want to review this PR? I think we should be OK to merge.

webbertakken:
First of all, amazing rewrite! 

The only comment I have is that ultimately we'll move to game-ci/cli. Which already contains a full rewrite (complete refactor of everything) as well.

It's no problem if the code bases diverge for me though. Just keep in mind that this repo will eventually use the cli under the hood and all code (apart from workflows) will effectively be obsolete at that point.

That said, you have my full support to move forward with this.

PR #558, opened 2023-08-20
Removed checks for old unity versions

EliasVal:
#### Changes

Changed referenced EOL versions in tests to LTS and Current versions (2018.x-2020.x -> 2021.x-2023.x)

#### Checklist

### DIFF
--- a/.github/workflows/build-tests-mac.yml
+++ b/.github/workflows/build-tests-mac.yml
@@ -20,15 +20,11 @@ jobs:
         projectPath:
           - test-project
         unityVersion:
-          - 2019.4.40f1 # Minimum version for IL2CPP
-          - 2020.1.17f1
-          - 2020.2.7f1
-          - 2020.3.44f1
-          - 2021.1.28f1
-          - 2021.2.19f1
-          - 2021.3.18f1
+          - 2021.3.29f1
           - 2022.1.24f1
-          - 2022.2.6f1
+          - 2022.2.21f1
+          - 2022.3.7f1
+          - 2023.1.9f1
         targetPlatform:
           - StandaloneOSX # Build a MacOS executable
 
--- a/.github/workflows/build-tests-ubuntu.yml
+++ b/.github/workflows/build-tests-ubuntu.yml
@@ -46,28 +46,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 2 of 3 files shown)


EliasVal:
@webbertakken As promised :)
Let me know if any further changes are needed

webbertakken:
Looks good to me. 

Note that this is just the checks to make sure `unity-builder` works as expected. We'll still need to do the same for `test-runner` and other actions that use the images. https://github.com/game-ci/docker/issues/220 is for changing the workflow to no longer even publish these images in the first place.

EliasVal:
> We'll still need to do the same for `test-runner` and other actions that use the images. 
@webbertakken So just change any reference to a version earlier than 2021.x to use 2022 LTS?

webbertakken:
> > We'll still need to do the same for `test-runner` and other actions that use the images.
> 
> @webbertakken So just change any reference to a version earlier than 2021.x to use 2022 LTS?

Not sure I understand the question to be honest. Also we're supporting all non-EOL versions, not just LTS.

So basically what we have to do is stop publishing EOL images in the `game-ci/docker` repository (which is done in `game-ci/versioning-backend` I believe. And the usages of it, like here in builder like you did.

webbertakken:
Which is to say you marked this PR as "resolves" the docker issue, but that is to stop publication of EOL images, not just usage. So we need to unlink that resolves-link.

EliasVal:
> Which is to say you marked this PR as "resolves" the docker issue, but that is to stop publication of EOL images, not just usage. So we need to unlink that resolves-link.

Sure, on it!

> Not sure I understand the question to be honest. Also we're supporting all non-EOL versions, not just LTS.
> 
> So basically what we have to do is stop publishing EOL images in the game-ci/docker repository (which is done in game-ci/versioning-backend I believe. And the usages of it, like here in builder like you did.


What I meant here is, to change every field that references the EOL versions to 2022.x (2021.x is EOL in April so 2022.x gives us more headroom)

i.e `game-ci/unity-test-runner/src/model/docker.test.ts`, `const image = '....2019.2.11f1...'`, to `const image = '...2022.3.7f1....'`

webbertakken:
I see. Yes sounds good to me!

webbertakken (approved):

Issue #340, opened 2022-02-20
Unable to import packages from private repo


justin-pierce:
**Bug description**

Have two packages hosted on github as private repos. 

Getting this error: 
```
An error occurred while resolving packages:
  Project has invalid dependencies:
    <package_id>: Error when executing git command. Failed to add the ECDSA host key for IP address '140.82.114.4' to the list of known hosts (/root/.ssh/known_hosts).
    ERROR: Repository not found.
    fatal: Could not read from remote repository.

    Please make sure you have the correct access rights
    and the repository exists.
```
(140.82.114.4 is a github.com ip)

**How to reproduce**

Yaml snippet: 
```
      # setup ssh agent to access private package repos in manifest
      - uses: webfactory/ssh-agent@v0.5.4
        with:
          ssh-private-key: |
            ${{ secrets.SSH_PRIVATE_KEY }}
            ${{ secrets.SSH_PRIVATE_KEY_TWO }}
            
      # test git clone (this successfully clones the repos, suggesting ssh-agent is setup properly)
      # - name: Git Clone Test
      #   run: |
      #     git clone git@github.com:<repo_1>.git
      #     git clone git@github.com:<repo_2>.git
      #     cd AREngine
      #     ls

      # Build
      - name: Build project
        uses: game-ci/unity-builder@v2
        env:
          UNITY_EMAIL: ${{ secrets.UNITY_EMAIL }}
          UNITY_PASSWORD: ${{ secrets.UNITY_PASSWORD }}
          UNITY_SERIAL: ${{ secrets.UNITY_SERIAL }}
        with:
          targetPlatform: iOS
          sshAgent: ${{ env.SSH_AUTH_SOCK }}
          buildMethod: ContinuousIntegrationLegacy.Build_iOS
```
Format in manifest is: `ssh://git@github.com/<company>/<project>.git`

**Expected behavior**

I expect it to successfully import these packages

**Additional details**

Followed these docs: https://game.ci/docs/github/builder#private-github-repositories
And ssh-agent docs: https://github.com/webfactory/ssh-agent

[...]

PR #568, opened 2023-09-06
Support multiple GitHub SSH deploy keys

eronnen:
This PR's goal is to solve [#340](https://github.com/game-ci/unity-builder/issues/340), hope it can be a good enough of a solution :)

#### Background 

The [webfactory/ssh-agent](https://github.com/webfactory/ssh-agent) action handles the issue of authenticating with multiple SSH deploy keys to Github by adding the multiple public keys to the `.ssh` folder with custom file names, and then using them accordingly by adding `url.*.insteadOf` git configs.

The problem is that these git configs as well as the SSH public keys are not uploaded to the container, so this mirroring doesn't work in the build.

#### Changes

- add `sshPublicKeysDirectoryPath` input, which is a custom `.ssh` that should contain the public keys and `config` file.
- pass a `GIT_CONFIG_EXTENSIONS` environment variable to the container, which is a list of `key=values` configs that can be retrieved from `git config --list` that the git inside the container will use.

#### Usage

With these changes, it's possible to fix the multiple SSH keys issue by using the following steps:

```yaml
- uses: webfactory/ssh-agent@v0.8.0
  with:
  ssh-private-key: |
    ${{ secrets.FIRST_KEY }}
    ${{ secrets.NEXT_KEY }}
    ${{ secrets.ANOTHER_KEY }}

- name: Prepare SSH config for unity builder
  run: |
    mkdir $HOME/.ssh_docker
    cp $HOME/.ssh/config $HOME/.ssh_docker/
    cp $HOME/.ssh/key* $HOME/.ssh_docker/
    cp $HOME/.ssh/known_hosts $HOME/.ssh_docker/
    sed -i 's/\/home\/runner/\/root/g' $HOME/.ssh_docker/config
    sudo chown -R root:root $HOME/.ssh_docker
    GIT_CONFIG_EXTENSIONS=$(git config --list | grep '^url\.')
    {
      echo 'GIT_CONFIG_EXTENSIONS<<EOF'
      echo "$GIT_CONFIG_EXTENSIONS"
      echo EOF
    } >> "$GITHUB_ENV"
  shell: bash

- uses: game-ci/unity-builder@v3
  ...
  with:
    sshAgent: ${{ env.SSH_AUTH_SOCK }}
    sshPublicKeysDirectoryPath: ${{ env.HOME }}/.ssh_docker
```

#### Checklist


src/model/docker.ts
```diff
@@ -89,6 +112,13 @@ class Docker {
             --volume "${actionFolder}/default-build-script":"c:/UnityBuilderAction" \
             --volume "${actionFolder}/platforms/windows":"c:/steps" \
             --volume "${actionFolder}/BlankProject":"c:/BlankProject" \
+            ${sshAgent ? `--volume ${sshAgent}:c:/ssh-agent` : ''} \
+                ${
+                  sshAgent && !sshPublicKeysDirectoryPath
+                    ? `--volume c:/Users/Administrator/.ssh/known_hosts:c:/root/.ssh/known_hosts`
```
webbertakken, line 118: Could you explain where `c:/root` comes from? Have you tested that this works?

webbertakken (approved):
The approach looks elegant enough if you ask me.

What do you think @GabLeRoux @davidmfinol @AndrewKahr?

src/model/docker.ts
eronnen: I tested only for linux runner, tried to port to windows as best as I can...
but looking at it again seems that `sshAgent` isn't used in Windows anyway so probably I should remove these windows changes enitrely

webbertakken: Yea better to have working software with missing features than the added complexity of a non working piece of which nobody knows how much of it was intended to work.

Thanks.

davidmfinol:
Looks good to me.
Just need to raise a PR to update the docs page as well: https://github.com/game-ci/documentation/blob/main/docs/03-github/04-builder.mdx

eronnen:
Thank you for the response!

added complementary documentation and `unity-test-runner` PRs:

* https://github.com/game-ci/unity-test-runner/pull/240
* https://github.com/game-ci/documentation/pull/416

Issue #442, opened 2022-09-21
Configuration option to suppress `-quit`


DecentM:
**Context**

I'm trying to use unity-builder to run a pre-existing build pipeline, that uses callbacks heavily internally. Because the initial step is also a callback based function, synchronously the function finishes execution while build preparations are still happening. Since the function I specified ends while there are still things happening, it causes Unity to quit because the `-quit` parameter is hardcoded into the build script. I believe I could still use unity-builder as-is if there was a way to suppress `-quit` and it'd let me just call `EditorApplication.Exit` when I know that the build process finished.

**Suggested solution**

Please add a new [configuration option](https://game.ci/docs/github/builder#configuration-options) to the GitHub Action that causes the `-quit` command line argument to not be passed to the editor.

**Considered alternatives**

If I understand correctly, Unity should treat returned coroutines as an indicator of the build's status, but in practice the editor quit before the build could start regardless when I tried. I also tried running a Task and having the main thread wait for in synchronously but as expected, that just hung the editor.

**Additional details**

Starting from [this line](https://github.com/game-ci/unity-builder/blob/3d0eb0805bbe2df90e319abbe42e41d0b9105625/dist/platforms/ubuntu/steps/build.sh#L128) in the build script, every other argument is customisable, except for `-quit`. Was it intended to be controllable at some point perhaps?

webbertakken:
Hey @DecentM, thanks for creating this issue and especially for your detailed description.

This is planned and can be tracked in https://github.com/game-ci/cli/issues/13 (the CLI will hold the core logic for unity-builder in v3.0.0).

DecentM:
Thanks for the quick response @webbertakken, that's great to know!  

Is there a way to use it right now instead of having to halt my progress on the feature I wanted to use this config option for? I can see that the issue is still open, so maybe it'd be worth it to backport it into v2 as it feels like a gap in the current functionality that doesn't depend on other changes you guys are planning for v3.

webbertakken:
Certainly. 

You can fork unity-builder and change entrypoint.sh and its sub-scripts inside the dist folder to do so.

PR #574, opened 2023-09-20
`manualExit` suppresses `-quit`, useful for buildMethods with async calls

tobyspark:
#### Changes

- Adds `with: manualExit` parameter to the action, that when present suppresses the Unity `-quit` command line argument. See https://github.com/game-ci/cli/issues/13 and https://github.com/game-ci/unity-builder/issues/442

#### Checklist

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -31,6 +31,10 @@ inputs:
     required: false
     default: ''
     description: 'Path to a Namespace.Class.StaticMethod to run to perform the build.'
+  manualExit:
+    required: false
+    default: ''
+    description: 'Suppresses `-quit`. Exit your build method using `EditorApplication.Exit(0)` instead.'
   customParameters:
     required: false
     default: ''
--- a/dist/platforms/ubuntu/steps/build.sh
+++ b/dist/platforms/ubuntu/steps/build.sh
@@ -119,7 +119,7 @@ echo ""
 
 unity-editor \
   -logfile /dev/stdout \
-  $( [ -n "${MANUAL_EXIT+set}" ] || echo "-quit" ) \
+  $( [ "${MANUAL_EXIT}" == "true" ] || echo "-quit" ) \
   -customBuildName "$BUILD_NAME" \
   -projectPath "$UNITY_PROJECT_PATH" \
   -buildTarget "$BUILD_TARGET" \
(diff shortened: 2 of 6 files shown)


tobyspark:
Notes
- ubuntu only so far. What else would need to be done?
- seems manualExit should be a boolean that defaults to false, but I didn’t see an example of that? Currently `build.sh` works on the presence of the env var, but the `javascript` requires a string for it to be passed through.

action.yml
```diff
@@ -31,6 +31,10 @@ inputs:
     required: false
     default: ''
     description: 'Path to a Namespace.Class.StaticMethod to run to perform the build.'
+  manualExit:
+    required: false
+    default: ''
+    description: 'Suppresses `-quit`. Exit your build method using `EditorApplication.Exit(0)` instead.'
```
webbertakken, line 37: Currently you're passing the variable directly to the bash script and doing complicated logic (I believe this is the first time we use `+set` in this repo) that differs from other `is set` checks in our bash scripts.

We want to move as much away from the bash scripts as possible (in favour of the higher level TypeScript code) and not pass arbitrary values to it. Normalisation and input should either be guaranteed to be correct or throw an error so that the user knows they've misconfigured something.

webbertakken (approved):
Looks good to me, mostly.

action.yml
tobyspark: It should default to false. Having it be the presence of a value wasn’t what I wanted, but was a first pass at using what I found; e.g. I didn’t see an example of boolean handling; `Input.getInput` returns a string. Happy to be pointed in the right direction.

Similarly, the bash script responds to the presence or not of the env var, which is clean from the POV of the separation of parsing the input. But I have no love for this kind of scripting and am happy to be pointed in the direction of how this this effect is achieved elsewhere in the codebase.

webbertakken: I'm in a bit of a hurry atm, but basically consider workflow inputs as input for the TS logic, and the TS logic to be responsible for passing in something deterministic to the bash script.

Ideally we'd remove the `string | undefined` in favour of `string` like the rest does, and parse that into a `true` or `false` or `1` or `0` value, and pass that into the script.

tobyspark:
Now models input as a boolean. I can squash the commit if that’s desired.

Reminder – haven’t touched the other platforms.

webbertakken (approved):

webbertakken:
This'll also need to be documented in https://github.com/game-ci/documentation

GabLeRoux (commented):
nice

Issue #572, opened 2023-09-13
Upgrading to builder@v3 broke iOS builds


mikekhristo:
**Bug description**

After upgrading to `Unity 2022.3.9f1` we had to upgrade to `unity-builder@v3` from v2. In doing so, builds started failing in the github workflow.

```
2023-09-13T05:02:58.7374030Z /Users/runner/work/_actions/game-ci/unity-builder/v3/webpack:/unity-builder/lib/model/exec-with-error-check.js:16
2023-09-13T05:02:58.7374970Z             throw new Error(`There was an error building the project. Please read the logs for details.`);
```

The `Build Results` show that while there was 1 error, the build actually suceeded:
```
2023-09-13T05:02:47.9154000Z ###########################
2023-09-13T05:02:47.9155020Z #      Build results      #
2023-09-13T05:02:47.9156470Z ###########################
2023-09-13T05:02:47.9158390Z 
2023-09-13T05:02:47.9158920Z Duration: 00:19:36.3575024
2023-09-13T05:02:47.9160530Z Warnings: 175
2023-09-13T05:02:47.9161560Z Errors: 1
2023-09-13T05:02:47.9162070Z Size: 1658650774 bytes
2023-09-13T05:02:47.9162340Z 
2023-09-13T05:02:47.9162350Z 
2023-09-13T05:02:47.9162530Z Build succeeded!
```

There is 1 `Exit Code 1` in the log output related to verifying the license. However, you can see that it was able to identify an active Pro license via serial number. 
```
2023-09-13T04:30:52.8443670Z [Licensing::Client] Error: Code 1 while verifying Licensing Client signature (process Id: 2878, path: "/Applications/Unity/Hub/Editor/2022.3.9f1/Unity.app/Contents/Frameworks/UnityLicensingClient.app/Contents/MacOS/Unity.Licensing.Client")
2023-09-13T04:30:52.8444810Z [Licensing::Module] Error: LicensingClient has failed validation; ignoring
2023-09-13T04:30:52.9150380Z [Licensing::Client] Handshaking with LicensingClient:
2023-09-13T04:30:52.9152020Z   Version:                 1.13.1+ca5f299
2023-09-13T04:30:52.9153460Z   Session Id:              f54d476e959547ba85e2435ca9303025
2023-09-13T04:30:52.9157890Z   Correlation Id:          11dba8cd44977f4ed2490be529ace13b
2023-09-13T04:30:52.9159640Z   External correlation Id: 328742284493430721
[...]

Issue #567, opened 2023-08-30
Workflow failing for android even after getting "Build Succeeded" 


Lokhandwalah:
**Bug description**

When I am trying to build unity for android platform, the build is successful but I still get the error and the workflow fails
I am using this in my current workflow

```
      - name: Build Unity
        uses: game-ci/unity-builder@v3
        env:
          UNITY_EMAIL: ${{ secrets.UNITY_USERNAME }}
          UNITY_PASSWORD: ${{ secrets.UNITY_PASSWORD }}
          UNITY_SERIAL: ${{ secrets.UNITY_PRO_SERIAL_KEY }}
        with:
          targetPlatform: Android
          projectPath: ./andromeda
          buildMethod: UnityBuilderAction.Build.DoBuildAndroidLibrary
          customImage: unityci/editor:ubuntu-2020.3.42f1-android-2
```

I can see this in the logs

```
Exiting batchmode successfully now!
Build succeeded

###########################
#       Build output      #
###########################
```

but this is what I get at the end

```
LICENSE SYSTEM [2023830 19:4:12] Received https://activation.unity3d.com/license.fcgi?CMD=3&TX=1696faac5bc1bc7a082b5da8501de02f&RX=GEN-7b86ec0d-4fa0-48c6-ab01-0d706921b358

LICENSE SYSTEM [2023830 19:4:12] Headers:
    HTTP/1.1 200 OK
    Cache-Control: no-cache, no-store, must-revalidate
    Pragma: no-cache
    Expires: 0
    Vary: Accept-Encoding
    Content-Encoding: deflate
    sc-request-id: 2706fd13-79c1-495b-a772-b525f74e6ee4
    Content-Type: text/xml
    Content-Length: 1132
    
LICENSE SYSTEM [2023830 19:4:12] Information: Unity license returned.
/github/workspace
Error: There was an error building the project. Please read the logs for details.
```

You can see the full logs [here](https://drive.google.com/file/d/11JIEt8-C90TUhpnXPhRiD6KlZ7iIQMNR/view?usp=sharing)

**Expected behavior**

I should be able to move forward in my workflow

**Additional details**

AndrewKahr:
See if v4 actions fix the issue, v3 had a different error-checking strategy which was a bit too aggressive

Issue #564, opened 2023-08-28
"Build succeeded" but still getting "There was an error..."


Sonorpearl:
**Bug description**

Hello lovely community,

I'm trying out GameCI on GitHub Actions with the ubuntu-latest runner on Unity version 2021.3.16f1 for building WebGL. My first step which it already fails on is building the Addressable content.

Everything seems fine, since I'm gettings logs like:
```
2023-08-28T16:29:20.2936686Z Addressable content successfully built (duration : 0:07:34.394)
2023-08-28T16:29:21.0915668Z Addressables: finished with override version: 16
2023-08-28T16:29:21.0924390Z Batchmode quit successfully invoked - shutting down!
2023-08-28T16:29:32.7973754Z Exiting batchmode successfully now!
2023-08-28T16:29:32.9231549Z Build succeeded
```

But then I'm still getting this:
```
2023-08-28T16:29:37.8266881Z ##[error]There was an error building the project. Please read the logs for details.
```

The only errors I see are:
```
2023-08-28T16:13:00.0946253Z ##[warning]Fetch --unshallow caught: Error: Failed to run "git fetch --unshallow".
 In-command error caught: Error: The process '/usr/bin/git' failed with exit code 1
```

and:
```
2023-08-28T16:15:23.4628679Z [LicensingClient] Channel doesn't exist: "LicenseClient-root"
2023-08-28T16:15:23.4633461Z [Licensing::Module] Successfully launched the LicensingClient (PId: 99)
2023-08-28T16:15:23.7765119Z [SignatureVerifier] Application signature verification not supported on this platform.
2023-08-28T16:15:23.8428377Z [LicensingClient] Handshaking with LicensingClient (version: 1.9.0+249add7)
2023-08-28T16:15:23.8429435Z [Licensing::Module] Successfully connected to LicensingClient on channel: "LicenseClient-root" (connect: 0.31s, validation: 0.07s, handshake: 0.00s)
2023-08-28T16:15:23.8429998Z [Licensing::Module] Connected to LicensingClient (PId: 99, launch time: 0.00, total connection time: 0.38s)
2023-08-28T16:15:23.8430493Z Entitlement-based licensing initiated
2023-08-28T16:15:23.8430999Z [Licensing::Module] Error: Access token is unavailable
2023-08-28T16:15:23.9672483Z [LicensingClient] Licenses updated successfully
```

and:
```
[...]

Issue #563, opened 2023-08-26
Disabling the check for the "Build results" output?


mikeage:
**Context**

We use a custom build action, and it doesn't print the same format of "Build results" that the default build method does. As a result, we're told that our build fails, even when it actually succeeds.

**Suggested solution**

Add an option to disable this check.

**Considered alternatives**

Adding this game-ci specific print to our build method.

**Additional details**

Raised on discord at https://discord.com/channels/710946343828455455/1143838569220280321

AndrewKahr:
Try updating to v4 and see if this is solved now

mikeage:
Unfortunately, v4 broke our flow for building from forks (where we had a license baked in to the github file, since secrets are inaccessible), and I'm not sure if the proper fix is to put a username/password there too. I've asked on the Discord (https://discord.com/channels/710946343828455455/1174438356755697784/1174438356755697784) and once this is resolved, I'll be able to see if our PR at https://github.com/icosa-foundation/open-brush/pull/557 works.

mikeage:
Indeed, this works now. Thanks

PR #575, opened 2023-09-20
`errorWhenMissingUnityBuildResults` exposed as `with` parameter

tobyspark:
#### Changes

Seemingly a breaking change in v2 → v3 is the action reporting error if the built player app isn’t present. Not all Unity builds are for the player, e.g. addressable content. Examining the build scripts shows a boolean `errorWhenMissingUnityBuildResults` managing this behaviour. This is now exposed as a `with` parameter in the action YAML. Also, a slightly opinionated refactor of `docker.run` to include an options object instead of the fixed signature requiring all options prior to `errorWhen...` to be defined.

#### Checklist

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -217,6 +217,12 @@ inputs:
     description:
       'The path to mount the workspace inside the docker container. For windows, leave out the drive letter. For example
       c:/github/workspace should be defined as /github/workspace'
+  errorWhenMissingUnityBuildResults:
+    default: true
+    required: false
+    description:
+      'Check if Unity build product is present after build, and error if not. Set to false to not check, useful if
+      producing alternative build products.'
 
 outputs:
   volume:
--- a/src/index.ts
+++ b/src/index.ts
@@ -25,7 +25,13 @@ async function runMain() {
       if (process.platform === 'darwin') {
         MacBuilder.run(actionFolder);
       } else {
-        await Docker.run(baseImage.toString(), { workspace, actionFolder, ...buildParameters });
+        await Docker.run(
+          baseImage.toString(),
+          { workspace, actionFolder, ...buildParameters },
+          {
+            errorWhenMissingUnityBuildResults: buildParameters.errorWhenMissingUnityBuildResults,
+          },
+        );
       }
     } else {
       await CloudRunner.run(buildParameters, baseImage.toString());
--- a/src/model/build-parameters.ts
+++ b/src/model/build-parameters.ts
@@ -84,6 +84,7 @@ class BuildParameters {
   public cacheUnityInstallationOnMac!: boolean;
   public unityHubVersionOnMac!: string;
   public dockerWorkspacePath!: string;
+  public errorWhenMissingUnityBuildResults!: boolean;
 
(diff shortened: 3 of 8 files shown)


action.yml
```diff
@@ -217,6 +217,12 @@ inputs:
     description:
       'The path to mount the workspace inside the docker container. For windows, leave out the drive letter. For example
       c:/github/workspace should be defined as /github/workspace'
+  errorWhenMissingUnityBuildResults:
+    default: true
+    required: false
+    description:
+      'Check if Unity build product is present after build, and error if not. Set to false to not check, useful if
+      producing alternative build products.'
```
webbertakken, line 225: This is effectively a discrepancy in the application (builder and/or container). I feel adding an option in the (functional) public api isn't the answer to this technical issue. 

We should probably agree on how the application should handle this case and not expose anything.

webbertakken (changes requested):
Unnecessary addition to public api that doesn't add any value.

tobyspark:
> This is effectively a discrepancy in the application (builder and/or container). I feel adding an option in the (functional) public api isn't the answer to this technical issue.

Is this a technical issue? There may indeed be some internal need to change this parameter depending on builder or container. But that’s not why I made the PR. There appears to be a design change in v3 that a successful action completion requires there to be a built player app. That means the Unity Builder action isn’t suitable for a whole range of uses, for example building addressables content. A quick look at the issues list shows many people hitting this, e.g. #563 #564 #567 #572

webbertakken:
You're probably right. And you're definitely more in the know than I am. 
Sorry if my review isn't accurate!

Perhaps @davidmfinol and @AndrewKahr could continue the review on this one?

webbertakken:
Also @GabLeRoux

tobyspark:
I’ve also just read the discord post mentioned in #563, which prompted me to re-read the code for `exec-with-error-check`. I’m not sure how I got “the existence of a built player app” from that, so apologies on that misdirected inaccuracy.

The basic point remains though – that it’s requiring something specific to be output by the build method, that not all build methods will output. This makes me wonder if a more elegant design might be either
- only perform the check if a custom `buildMethod` is _not_ specified in the workflow
- document a simpler string that will be checked for before trying the highly specific regex.

webbertakken:
I think one of the problems we're facing of why we're checking for a specific text in the first place is that the exit code isn't always reliable. But it is a long time ago that I've implemented the first code for that and unfortunately I've lost context.

tobyspark:
> document a simpler string that will be checked for before trying the highly specific regex.

Separately, I’ve hit the issue that you can’t write a `GITHUB_STEP_SUMMARY` from one’s Unity code, as it’s running inside a container. Putting the motivation behind this PR together with that, and 
- we could write out the Unity build results as a step summary from `exec-with-error-check.ts`
- and we’d also check for a string in `stdout` like `GAME_CI_STEP_SUMMARY`
If either pass, we don’t error.

tobyspark:
I’m going to abandon this PR in favour of the approach above. It’s more elegant, solves more problems, and having implemented it for my own purposes, can be self-documenting for someone who experiences the conundrum of seeing ‘Build succeeded’ (i.e. exit code 0) but have the step fail on the results check.

Issue #542, opened 2023-06-20
Github actions hosted runner fails for WSAPlayer with `windows-2022`


camnewnham:
**Bug description**

Building WSAPlayer using github actions  with a hosted runner fetches an incompatible docker image and produces the error: `The container operating system does not match the host operating system.`

This happens with `windows-2022`. `windows-2019` works as expected.

**How to reproduce**

- Create a github actions workflow with `runs-on: windows-latest`
- Attempt to build a `WSAPlayer` project

**Expected behavior**

Unity project is built.

**Additional details**

Github hosted runner details:
```
Current runner version: '2.305.0'
Operating System
  Microsoft Windows Server 2022
  10.0.20348
  Datacenter
Runner Image
  Image: windows-2022
  Version: 20230612.1
  Included Software: https://github.com/actions/runner-images/blob/win22/20230612.1/images/win/Windows2022-Readme.md
  Image Release: https://github.com/actions/runner-images/releases/tag/win22%2F20230612.1
```

Error in build step:
```
Status: Downloaded newer image for unityci/editor:windows-2022.3.1f1-universal-windows-platform-1
docker: Error response from daemon: hcs::CreateComputeSystem 57f9bea707270d54c11b52dc1b4a88087228987678ecf821e16279ce12abacc3: The container operating system does not match the host operating system.
Error: The process 'C:\Windows\system32\docker.exe' failed with exit code 125
```

davidmfinol:
This issue is dependent on updating the docker images: https://github.com/game-ci/docker/issues/213

AndrewKahr:
V4 is now released and uses windows-2022

Issue #540, opened 2023-06-05
Allow us to set Docker's Memory Usage


GitFlip:
**Context**

Within the `docker.ts` file there are two methods: `getLinuxCommand` and `getWindowsCommand` that return the `docker run` command string. There is no way for us to set the memory usage of this docker container that runs.

**Suggested solution**

Add `-m` argument to `docker run` command that takes an input from the Builder's Configuration Options. For example:

```yml
- uses: game-ci/unity-builder@v2
  with:
    dockerMemory: 4GB
```

**Considered alternatives**

Did not consider alternatives

**Additional details**

The reason for this is because when using a GitHub hosted Runner the VM has 7GB of RAM, and Unity Builder seems to think all of it is available to it - see snippet from Builder's log:
```
###########################
#    Building project     #
###########################

Unity Editor version:    2021.3.6f1 (7da38d85baf6)
Branch:                  2021.3/staging
Build type:              Release
Batch mode:              YES
System name:             Linux
Distro version:          #45-Ubuntu SMP Mon Apr 24 15:40:42 UTC 2023
Kernel version:          5.15.0-1038-azure
Architecture:            x86_64
Available memory:        6943 MB
```

The docker container however does not have that much memory and this causes the container to run out of memory when building giving an `Build failed, with exit code 137` because Unity Builder is trying to use more memory than what has been allocated to the container.

Adding the ability to set the memory allocation for the docker container will allow us to increase it to match what the Builder thinks is actually available.

shubhank008:
Anything on this ?

webbertakken:
Not sure if the proposed solution will work but we're always open to contributions! Counting on the community for this enhancement.

wilg:
I mentioned this a while ago here https://github.com/game-ci/unity-builder/issues/432

ejnarvala:
We are also seeing this issue when using our kubernetes ARSS GHA runners to process jobs. As part of this setup, we are using a docker-in-docker setup where each runner pod has a runner container and a DinD container with daemon exposed to the runner.

What we see is a similar issue where the unity app seems to be reporting the total memory of the VM/node that it's running in rather than the request/limit set for the DinD container. This leads to frequent OOMs (exit code 137). 

We have tried using the `dockerCpuLimit` and `dockerMemoryLimit` parameters, but this didn't seem to help. It only served to fix an invalid argument issue where not setting these values would result in this error:
```
[...]

Issue #439, opened 2022-09-20
Powershell never exits on Windows and the build hangs forever


wilg:
I don't know if it something with my configuration of a local runner, but something is preventing the Docker container from terminating. I think it is that Powershell never exits after completing the build script, even when explicitly exited. I've been taking some stabs in the dark on my fork, but haven't had any success here. https://github.com/wilg/unity-builder

webbertakken:
Would you mind attaching your workflow file, so we can see your configuration?
Could you also tell us what kind of runner you're using?

wilg:
I'm using local runners, one on Mac, one on Windows. My idea is to have the Mac build Mac IL2CPP and have the Windows machine build Windows IL2CPP and Linux IL2CPP.

The Mac build works properly (though I haven't gotten Steam uploads working yet because of https://github.com/game-ci/steam-deploy/issues/35).

The Windows machine is Windows 11 Pro. I installed Docker via Docker Desktop, maybe that's at issue.

The Windows build fully completes, but the Docker instance hangs. It looks like the Powershell script completes (no code is called after the manual exit I added). However I see that it's possible for Powershell to hang even in this case. https://stackoverflow.com/questions/2041799/powershell-script-gets-stuck-doesnt-exit-when-called-from-batch-file

I have some puzzling configuration issues with memory with docker on Windows, documented here, but manually setting memory/cpu flags got it working. https://github.com/game-ci/unity-builder/issues/432

If I manually terminate the container from Docker Desktop, `docker run` exits (though with a crazy status code of what appears to be INT_MAX) and the build proceeds as normal.


```
name: Build Players
on:
  workflow_dispatch:
    inputs:
      build_mac:
        description: "Build for macOS"
        required: true
        type: boolean
      build_windows:
        description: "Build for Windows"
        required: true
        type: boolean
      build_linux:
        description: "Build for Linux"
        required: true
        type: boolean
env:
  NODE_TLS_REJECT_UNAUTHORIZED: 0
jobs:
  windows_host:
    if: inputs.build_linux || inputs.build_windows
    name: Build for ${{ matrix.platform[1] }} on ${{ matrix.platform[0] }}
    runs-on: [self-hosted, "${{ matrix.platform[0] }}"]
    strategy:
      fail-fast: false
      matrix:
        unityVersion:
          - 2021.3.9f1
        platform:
          - [Windows, StandaloneWindows64]
    steps:
      - uses: actions/checkout@v3
        name: Git Checkout + LFS
[...]

Issue #432, opened 2022-09-11
Cannot configure Docker to use more than 1GB of memory when running a Windows build on Windows


wilg:
Absolutely losing my mind trying to figure out why Docker on Windows has a 1GB memory limit that won't allow my app to build. This seems to be the issue: https://stackoverflow.com/questions/43460770/docker-windows-container-memory-limit but none of the workarounds seem to work properly. I think passing a `-m` flag or whatever into the usage of Docker from unity-builder might work, but I have no idea. Any help is greatly appreciated. Using a local action runner on Windows 11 Pro.

wilg:
I forked it to allow me to just pass memory and cpu count vars, since I could not figure out how to change the defaults. https://github.com/wilg/unity-builder

AndrewKahr:
Try updating to v4 and see if this is solved now. It should automatically pick a higher limit now but you can manually override the value too.

PR #588, opened 2023-10-27
Feature/windows upgrades

AndrewKahr:
#### Changes

- Allow updating container memory and cpu limits for Windows. Previously, they defaulted to 1cpu and 1gb ram which was far too low and it seems docker wouldn't allocate all available resources. Now it will use all available cores and 80% of system memory.
- Allow setting docker isolation mode for windows. Defaults to default to ensure behavior doesn't change from prior versions but now you can do stuff like force process mode on non-server versions which grants a performance uplift during runs
- Added logic to allow building Android on Windows. Android doesn't support burst when built on Linux, only on Windows and macOS. Thus we need to allow building Android on WIndows due to the major performance benefits of Burst.
- Support Windows 2022 and VS2022 by mounting the x64 Visual Studio path in addition to the x86 path to maintain compatibility with VS2019 and older
- Attempted fixes for windows builds hanging by killing the regsvr32 process after registering VS dll and using a different method to launch Unity. Unsure if this is a definite fix so I am leaving in several debug calls to print out running processes so we have more data to work with on chasing down this bug. I suspect there's a process that's hanging around that isn't cleaning itself up or is getting into some kind of deadlock situation and needs to be killed. But the changes I've made have seen no hangs on building during docker test workflows when previously there would be at least 3-5 hanging builds.

#### Related PRs
This PR also needs game-ci/documentation#433 to be merged to document the new parameter inputs. Then we will be able to work on merging a PR for the updated docker images to go along with this PR

#### Related Issues
* #540 
* #439
* #432 
* #542

#### Checklist

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -117,6 +117,11 @@ inputs:
       'Amount of memory to assign the docker container. Defaults to 95% of total system memory rounded down to the
       nearest megabyte on Linux and 80% on Windows. On unrecognized platforms, defaults to 75% of total system memory.
       To manually specify a value, use the format <number><unit>, where unit is either m or g. ie: 512m = 512 megabytes'
+  dockerIsolationMode:
+    required: false
+    default: 'process'
+    description:
+      'Isolation mode to use for the docker container. Can be either process or hyperv. Only applicable on Windows'
   allowDirtyBuild:
     required: false
     default: ''
--- a/action.yml
+++ b/action.yml
@@ -121,8 +121,9 @@ inputs:
     required: false
     default: 'default'
     description:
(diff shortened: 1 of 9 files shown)


action.yml
```diff
@@ -119,9 +119,10 @@ inputs:
       To manually specify a value, use the format <number><unit>, where unit is either m or g. ie: 512m = 512 megabytes'
   dockerIsolationMode:
     required: false
-    default: 'process'
+    default: 'default'
```
webbertakken, line 122: Do you think we should describe in the comment what `default` does in practice?

webbertakken (approved):

PR #590, opened 2023-11-07
License Activation fixes and Github Annotations

AndrewKahr:
#### Changes

- Fix license activation issues
- Add Github Actions annotations to default build script to make errors and warnings from Unity easier to find and diagnose
- Add error annotation for license activation error
- Update out-of-date node packages
- Further serial masking improvements
- Update test project and workflows to improve stability and relevance

#### Related PRs

- #576 - added baseline to aid with this PR
- https://github.com/game-ci/documentation/pull/433 - Updated docs for activation changes

#### Checklist


AndrewKahr:
iOS build issue seems to be a Unity issue and not from the changes to the action

.github/workflows/build-tests-mac.yml
```diff
@@ -20,11 +18,12 @@ jobs:
         projectPath:
           - test-project
         unityVersion:
-          - 2021.3.31f1
-          - 2022.3.11f1
-          - 2023.1.17f1
+          - 2021.3.32f1
+          - 2022.3.13f1
+          - 2023.1.19f1
```
webbertakken, line 23: I think this is fine for now, but I don't think it should be a goal to have these updated to the latest all the time, as there's value in having reference versions. Especially for long running issues.

dist/platforms/ubuntu/steps/activate.sh
```diff
@@ -4,76 +4,30 @@
...
 
   # Activate license
   unity-editor \
     -logFile /dev/stdout \
     -quit \
     -serial "$UNITY_SERIAL" \
     -username "$UNITY_EMAIL" \
-    -password "$UNITY_PASSWORD"
+    -password "$UNITY_PASSWORD" \
+    -projectPath "/BlankProject"
```
webbertakken, line 22: Are you sure about this change? Does it work for all versions? We separated between the two methods because of necessity back then. A single flag can make a rigorous difference in the Unity command line and show entirely different behaviour at times.

If you're sure about this, I welcome the simplification :)

webbertakken (approved):

dist/platforms/ubuntu/steps/activate.sh
AndrewKahr: Yea so the reason I switched to this is it seems that activating with serial likes having a real project path provided. I was getting some odd errors without it. I think when it was doing license file activation it didn't care since it was probably just moving the ulf to the right folder rather than actually spinning up Unity for the licensing client which is what happens now since manual ulf doesn't seem to work anymore unfortunately.

GabLeRoux (commented):
nice work here 🙌

Issue #597, opened 2023-11-16
Retry on license failures


asbjornu:
**Context**
Something like 10-20% of our builds fail due to not being able to obtain a license from Unity. These failures seem to be intermittent and a re-run of the build usually "fixes" the problem. 

**Suggested solution**
It would be nice if the `game-ci/unity-builder` action was able to detect a license failure and retry N times before failing the entire build.

**Considered alternatives**
We could use something like [action::retry](https://github.com/marketplace/actions/retry-action) to retry the `game-ci/unity-builder` action on failure, but that would not take into consideration why the action failed, and would retry build failures that are due to faults of our own (and need us to change stuff in order to succeed, making the retry unnecessary).

**Additional details**
The build log for the last license failure looks like this (identifiers abbreviated with `…` due to privacy; no idea whether this is needed or not, but better to fail on the safe side, imho):

```log
[Licensing::Module] Trying to connect to existing licensing client channel...
[Licensing::IpcConnector] Connection attempt to the License Client on channel: "LicenseClient-root" failed because channel doesn't exist; code: "0x80000002"
[Licensing::Module] Successfully launched the LicensingClient (PId: 150)
[Licensing::IpcConnector] Successfully connected to the License Client on channel: "LicenseClient-root"
[SignatureVerifier] Application signature verification not supported on this platform.
[Licensing::Client] Handshaking with LicensingClient:
  Version:                 1.12.1+9338cad
  Session Id:              3ebc…8b34
  Correlation Id:          6ede…aecf
  External correlation Id: 4320…8029
  Machine Id:              D7nT…YdM=
[Licensing::Module] Successfully connected to LicensingClient on channel: "LicenseClient-root" (connect: 0.62s, validation: 0.12s, handshake: 0.00s)
[Licensing::IpcConnector] Successfully connected to the License Notification on channel: "LicenseClient-root-notifications"
[Licensing::Module] Connected to LicensingClient (PId: 150, launch time: 0.00, total connection time: 0.74s)
[Licensing::Module] Error: Access token is unavailable; failed to update
[Licensing::Module] Error: Failed to return entitlement license
[Licensing::Client] Successfully queried for the EntitlementGroupsDetails
[Licensing::Module] Error: Serial number unavailable for ULF return; aborting operation
```

eronnen:
wouldn't it be better to perform the retry inside the docker?

mikeage:
[...]

Issue #521, opened 2023-03-07
some files and directories are owned by root:root and github runner user is not able to delete them :: self-hosted runner


davi020:
# Build the Unity WebGL build using game-ci/unity-builder@v2
      - name: Build WebGL
        uses: game-ci/unity-builder@v2
        env:
          UNITY_LICENSE: ${{ secrets.UNITY_LICENSE }}
          UNITY_EMAIL: ${{ secrets.UNITY_EMAIL }}
          UNITY_PASSWORD: ${{ secrets.UNITY_PASSWORD }}
        with:
          targetPlatform: WebGL
          buildName: WebGLBuild
          buildsPath: builds/webgl
          chownFilesTo: username:group (Tried with githib runner userid and group id as well)
          
      even I tried to use the chownFilesTo: username:group in my workflow file, still some files and directories are created with root:root permissions. 
      
      Directory names - builds,Library,Logs

Any help to fix this?

AndrewKahr:
Is it just he build the output folder or other folders too?

davi020:
@AndrewKahr  - These are the files that are owned by root in self-hosted runner after this action - 

[root@xxxxxxx]# ll | grep root
drwxr-xr-x.  2 root      root       6 Mar  7 09:26 _activate-license~
drwxr-xr-x.  3 root      root      19 Mar  7 09:26 builds
drwxr-xr-x.  9 root      root    4096 Mar  7 09:27 Library
drwxr-xr-x.  2 root      root       6 Mar  7 09:26 Logs
drwxr-xr-x.  2 root      root   12288 Mar  7 09:28 Temp
[root@xxxxxxxxx]#

AndrewKahr:
@davi020 It's interesting that _activate-license~ is there as the build scripts delete that folder after the build. Is this after a successful build or after a failed build?

davi020:
@AndrewKahr - Due to root:root permissions none of our workflows are not yet successful. The _activate-license~ is not getting deleted right after the build phase as well.

davi020:
@AndrewKahr  - Question - Why chownFilesTo is not working as expected here?

AndrewKahr:
That's what I'm trying to work out as we have code for that already:
```
# Make a given user owner of all artifacts
if [[ -n "$CHOWN_FILES_TO" ]]; then
  chown -R "$CHOWN_FILES_TO" "$BUILD_PATH_FULL"
[...]

Issue #504, opened 2023-02-10
Action doesn't change files permission on docker fail


MrFastwind:
**Bug description**

if unity fails with fatal signal, and docker fails, the directory remain with root permissions, and it doesn't change it back nor set it to *chownFilesTo*  variable.

**How to reproduce**
It's difficult, in my case was a an error with PlayerSettings.asset.
Others:
- set *chownFilesTo*
- run without being root
- have docker permissions

**Expected behavior**

even if it fails, the directory permission should change to *chownFilesTo* if present.
so it's possible to delete it without root permissions.

**Additional details**
log
```bash
Scripts have compiler errors.

Aborting batchmode due to failure:
Scripts have compiler errors.

./External/baselib/builds/Include/Internal/../C/Internal/Baselib_Semaphore_FutexBased.inl.h(151): Assertion failed (count >= 0) - Destruction is not allowed when there are still threads waiting on the semaphore.
Caught fatal signal - signo:5 code:128 errno:0 addr:(nil)
Trace/breakpoint trap (core dumped)
Exit code was: 133
Error: The process '/usr/bin/docker' failed with exit code 133
```

also it seams that docker code 133 means that it catched a SIGTRAP (i suppose from unity)

miguelvieirart:
Had problems with that, but managed to fix using a step that runs on failure to run chown on current directory

- name: Chown files on failure
        if: failure()
        run: |
          chown -R user:group ./

MrFastwind:
> Had problems with that, but managed to fix using a step that runs on failure to run chown on current directory
> 
> - name: Chown files on failure
>         if: failure()
>         run: |
>           chown -R user:group ./
> 

Thank you, but it would only work if the runner had root permission or sudo, and i don't have it in my environment

I would like to avoid spining up a container (that have root permission) and change owner from it.

Why unity is so peeky on permission?

Well, i'll probably going to close the issue my self in the next days

AndrewKahr:
v4 now has a `runAsHostUser` parameter to have the docker container not run as root which should fix these permission issues. If you run on a self hosted runner, just set it to true.

PR #600, opened 2023-11-21
Allow Running Container as Runner Host User

AndrewKahr:
#### Changes

- Added `runAsHostUser` to allow running the container as the same user as the host system. This fixes most permissions issues on self-hosted runners.
- Perform android sdk setup during entrypoint.sh to ensure it has root permissions if the user switches to a non-root user
- Automatically detect android sdk target version if parameters are not already provided to configure the sdk
- Generate a new uuid for machineID to ensure separate containers are unique to reduce license activation errors
- Add exponential retry strategy for Ubuntu license activations

#### Related PR
- https://github.com/game-ci/documentation/pull/438

#### Related Issues
- #597 
- #504
- #521

#### Checklist


davidmfinol (approved):

dist/platforms/ubuntu/entrypoint.sh
```diff
@@ -1,47 +1,63 @@
...
+
+groupadd -g $GROUPID $GROUPNAME
+useradd -u $USERID -g $GROUPID $USERNAME
+usermod -aG $GROUPNAME $USERNAME
+mkdir -p "/home/$USERNAME"
+chown $USERNAME:$GROUPNAME "/home/$USERNAME"
+
+# Normally need root permissions to access when using su
+chmod 777 /dev/stdout
+chmod 777 /dev/stderr
```
GabLeRoux, line 19: We used to have `chmod -R 777 "/BlankProject"` here. I don't have much context, but there might be security concerns about these two linces. Why is-this necessary exactly? Could it be more specific or behind a condition?

dist/platforms/ubuntu/entrypoint.sh
```diff
@@ -1,47 +1,63 @@
...
+USERNAME=$(stat -c '%U' "$fullProjectPath")
+USERID=$(stat -c '%u' "$fullProjectPath")
+GROUPNAME=$(stat -c '%G' "$fullProjectPath")
+GROUPID=$(stat -c '%g' "$fullProjectPath")
+
+groupadd -g $GROUPID $GROUPNAME
+useradd -u $USERID -g $GROUPID $USERNAME
+usermod -aG $GROUPNAME $USERNAME
+mkdir -p "/home/$USERNAME"
+chown $USERNAME:$GROUPNAME "/home/$USERNAME"
```
GabLeRoux, line 15: Since the script doesn't start with `set -x`, shoud we add some error handling for these 5 lines here?

dist/platforms/ubuntu/steps/runsteps.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
+#
+# Run steps
+#
+source /steps/set_extra_git_configs.sh
+source /steps/set_gitcredential.sh
+source /steps/activate.sh
+source /steps/build.sh
+source /steps/return_license.sh
```
GabLeRoux, line 10: Thinking out loud: Using source here might not respect the shebangs in the scripts, but so far, everything uses bash so I think this is fine.

GabLeRoux (commented):
I haven't tried this, it looks fine to me, but I left a few comments. I'm not exactly sure why we're changing permissions to stdout and stderr. I assume this does indeed solve issues with self-hosted runner, but I'm wondering if this should always be done or if it should be done only based on configuration 🤔

dist/platforms/ubuntu/entrypoint.sh
AndrewKahr: Good point, will add some error checking. Also realized I can put this inside the later if statement so it only runs when using the run as host option

dist/platforms/ubuntu/entrypoint.sh
AndrewKahr: The chmod didn't seem to fix the issue completely as a cancelled build wouldn't reach the line and cause runner agents to fail to delete the files. This PR is an overarching fix for permissions such that in the future if new files or folders are generated they will have the right ownership and we won't need to keep chasing these issues down. I'm not concerned on the security front because this is within the container and previously it was running as root already so any semblance of security was already out the window as the user could do anything they wanted to. These needed to be added so the non root user can write to stdout which is baked into the docker images for xvfb to work hence doing the workaround here. I will however move this within the if statement as it is only needed for non-root users.

dist/platforms/ubuntu/steps/runsteps.sh
GabLeRoux: Agreed, let's keep it this way as things are working fine. Extracting scripts like this is at least way more easier to read. Nicely done 👍

GabLeRoux (approved):

dist/platforms/ubuntu/entrypoint.sh
GabLeRoux: I wrote `set -x`, but I meant `set -e`, I edited my previous comment. I see this is resolved in [91ab33e](https://github.com/game-ci/unity-builder/pull/600/commits/91ab33e43b95154901ff0a03dfeb57db3f948ea5)

PR #630, opened 2024-02-19
Hotfixes for 4.2.0

AndrewKahr:
#### Changes

- Fix skipActivation not passed to macos ENV
- Fix linux il2cpp image tag for non-cloudrunner execution

#### Checklist


src/model/image-tag.ts
```diff
@@ -100,8 +109,12 @@ class ImageTag {
...
         // Unity versions before 2019.3 do not support il2cpp
-        if (process.env.USE_IL2CPP === 'true' && (major >= 2020 || (major === 2019 && minor >= 3))) {
-          return linuxIl2cpp;
+        if (major >= 2020 || (major === 2019 && minor >= 3)) {
+          if (providerStrategy === 'local') {
+            return linuxIl2cpp;
+          } else {
+            return process.env.USE_IL2CPP === 'true' ? linuxIl2cpp : linux;
+          }
         }
```
GabLeRoux, line 118: nitpick: I asked myself if this would be easier to read this way:

```suggestion
        const isVersionEligible = major > 2019 || (major === 2019 && minor >= 3);
        const shouldUseIl2cpp = providerStrategy === 'local' || process.env.USE_IL2CPP === 'true';
        
        if (isVersionEligible) {
          return shouldUseIl2cpp ? linuxIl2cpp : linux;
        } else {
          return linux;
        }
```

but after comparing this suggestion, I think tu current code is still better and in line with other cases from that switch case 👍  Looks good to me.

src/model/platform-setup/setup-mac.ts
```diff
@@ -168,6 +168,7 @@ class SetupMac {
     process.env.UNITY_VERSION = buildParameters.editorVersion;
     process.env.UNITY_SERIAL = buildParameters.unitySerial;
     process.env.UNITY_LICENSING_SERVER = buildParameters.unityLicensingServer;
+    process.env.SKIP_ACTIVATION = buildParameters.skipActivation;
```
GabLeRoux, line 171: This does indeed fix missing build param for macOS. This made me think; should we document this parameter in the docs?

GabLeRoux (approved):
Looks good to me 👍 Nice one.

I see there are some jobs failing. Not sure if it's related to current changes or if they're just flaky jobs 🤷

src/model/platform-setup/setup-mac.ts
AndrewKahr: Indeed we should, there's a PR up on the docs repo, just need an approval on the PR

AndrewKahr: PR Link: https://github.com/game-ci/documentation/pull/452

AndrewKahr:
> Looks good to me 👍 Nice one.
> 
> I see there are some jobs failing. Not sure if it's related to current changes or if they're just flaky jobs 🤷
> 
> Saving job output here in case job logs expire:
> 
> Details

Ah, it looks like switching tests to pure il2cpp was a bit too greedy on tests as Linux can't build Windows il2cpp. I think what I'll do is still merge this PR and get the hotfix out to unbreak people's pipelines then fix the test workflow since that won't impact anyone else using the action.

GabLeRoux:
> Ah, it looks like switching tests to pure il2cpp was a bit too greedy on tests as Linux can't build Windows il2cpp. I think what I'll do is still merge this PR and get the hotfix out to unbreak people's pipelines then fix the test workflow since that won't impact anyone else using the action.

Sounds good 🚀

PR #640, opened 2024-03-17
Hotfix: Fix version checking in image-tag

AndrewKahr:
#### Changes

- Update version checking in image-tag.ts to support 6000+ Unity versions.

#### Checklist

### DIFF
--- a/src/model/image-tag.test.ts
+++ b/src/model/image-tag.test.ts
@@ -2,7 +2,7 @@ import ImageTag from './image-tag';
 
 describe('ImageTag', () => {
   const testImageParameters = {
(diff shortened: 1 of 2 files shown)


webbertakken:
Not sure I agree with the reasoning on this one. Just because they add a new version scheme that's actually not so very different doesn't meant we have to make logic more implicit and call it out of scope. In the other PRs, we also still use a regex. The argument "if they change things again" is generally not a great one in software development either.

I'd prefer to keep specific rather than implicit and not architecting for things that might never be.

webbertakken (approved):

Issue #643, opened 2024-04-02
Building on self-hosted Windows 11: dockerWorkspacePath doesn't update --workdir


ah1053:
**Bug description**
Running Unity Builder on Windows 11 self-hosted results in this error `docker: Error response from daemon: the working directory 'c:/github/workspace' is invalid, it needs to be an absolute path.` 

Passing in dockerWorkspacePath as prompted doesn't seem to have an effect

Any help is appreciated

**How to reproduce**

- Setup a self-hosted runner on Windows 11 and run unity builder with default args but 

```
game-ci/unity-builder@v4
  with:
    targetPlatform: Android
    dockerWorkspacePath: /github/workspace
    unityVersion: auto
    versioning: Semantic
    androidExportType: androidPackage
    androidSymbolType: none
    runAsHostUser: false
    dockerIsolationMode: default
    containerRegistryRepository: unityci/editor
    containerRegistryImageVersion: [3](https://github.com/storyup/healium-xr-oculus-new/actions/runs/8528284125/job/23361561125#step:4:3)
    awsStackName: game-ci
    providerStrategy: local
    kubeVolumeSize: [5](https://github.com/storyup/healium-xr-oculus-new/actions/runs/8528284125/job/23361561125#step:4:5)Gi
    watchToEnd: true
    cacheUnityInstallationOnMac: false
    skipActivation: false
  env:
    UNITY_SERIAL: ***
    UNITY_EMAIL: ***
    UNITY_PASSWORD: ***
   ```

**Expected behavior**

Adding dockerWorkspacePath will override the `--workdir` however it doesn't

```
"C:\Program Files\Docker\Docker\resources\bin\docker.exe" run 
--workdir c:/github/workspace --rm 
--env UNITY_EMAIL=*** --env UNITY_PASSWORD=*** 
[...]

Issue #663, opened 2024-08-11
Unity Path Error on self-hosted Windows Runner


DavidJakszta:
**Bug description**

When building with game-ci/unity-builder@v4 and containerRegistryImageVersion: 2 the Unity Path isn't found.
The Unity Path either doesn't seem to be correct or something failed in the pulling/creating of the container. Also the action doesn't fail but that is of less concern to me at the moment.
![image](https://github.com/user-attachments/assets/72e6cacb-cdc4-40c0-9cb7-87198c778907)

**How to reproduce**

- Use unity-builder action in workflow on a self-hosted windows 10, with container RegistryImageVersion: 2 specified.
- Docker-Image: unityci/editor:windows-2021.3.22f1-windows-il2cpp-2
- Workflow yaml:
```
      - name: Build project
        uses: game-ci/unity-builder@v4
        env:
          UNITY_LICENSE: ${{ secrets.UNITY_LICENSE }}
          UNITY_EMAIL: ${{ secrets.UNITY_EMAIL }}
          UNITY_PASSWORD: ${{ secrets.UNITY_PASSWORD }}
        with:
          targetPlatform: StandaloneWindows
          buildMethod: BuildScript.PerformWindowsBuild${{ inputs.build-method }}
          #runAsHostUser: true
          dockerWorkspacePath: /github/workspace
          containerRegistryImageVersion: 2
```
**Expected behavior**

- The Unity installation is found and the license is activated succesfully.
**Additional details**
Rather a question: is it possible to ssh into the container or something to see what could be wrong with the path?

frostebite:
Heads up: v6.0.0-beta.1 (see #845) delegates the build invocation, platform setup, and licensing logic entirely to game-ci/cli's extracted implementation, replacing this repo's own `src/model/` code. That's a strong candidate to have changed the behavior reported here.

Leaving this open for now — once the thin wrapper merges, please retest against the new version and let us know if this is still reproducible. If so, we'll want to track it against game-ci/cli directly going forward.

frostebite:
Closing for now since the underlying implementation this reported against is being replaced by the thin-wrapper migration. Please reopen (or file a fresh issue) if this still reproduces once you're on the new version — happy to pick it back up.

PR #668, opened 2024-09-14
Fixed bug with additional 'c:' prefix for --workDir

Mannilie:
#### Changes

- Removed the `'c:'` prefix from the `--workDir` path for Windows hosts.

#### Related Issues

- https://github.com/game-ci/unity-builder/issues/643

repo.

#### Checklist

- [X] Docs (If new inputs or outputs have been added or changes to behavior that should be documented. Please make a PR
- [X] Readme (updated or not needed)
- [X] Tests (added, updated or not needed)

## Summary by CodeRabbit

- **New Features**
	- Improved compatibility of the Docker command for cross-platform usage by removing the hardcoded `C:` prefix from the workspace path.

- **Bug Fixes**
	- Enhanced flexibility in workspace path configuration, allowing for better adaptability across different environments.

### DIFF
--- a/src/model/docker.ts
+++ b/src/model/docker.ts
@@ -100,7 +100,7 @@ class Docker {
     } = parameters;
 
     return `docker run \
-            --workdir c:${dockerWorkspacePath} \
+            --workdir ${dockerWorkspacePath} \
             --rm \
             ${ImageEnvironmentFactory.getEnvVarString(parameters)} \
             --env GITHUB_WORKSPACE=c:${dockerWorkspacePath} \


AndrewKahr:
What seems a little odd about this is in v2, it was hardcoded to c:/github/workspace which nobody seemed to have any issues with. We currently have a Windows 11 runner building successfully on V4 without touching the workdir variable, so maybe there's another variable at play?

Here's the v3 commit with the changes for reference: https://github.com/game-ci/unity-builder/commit/3032a4ab97a9bb3fe204eeecd61b5abcc14885d3

Mannilie:
Hey @AndrewKahr, I agree. 
It does seem weird but as I understand, the path should be relative as opposed to absolute for the `--workDir` argument. 

Unless I'm missing something, even in the action items you've written, I believe we're on the same page that it should definitely be omitted from the command.
image

AndrewKahr:
It was instructed to leave out the drive letter because the variable was shared between Linux and Windows, so the Windows function would prepend the `c:` portion while the Linux one doesn't prepend the `c:`. So at the action level, it is still technically absolute pathing with an assumption of a c drive in the container for Windows containers. And at the docker level, it is supposed to be absolute.

This sounds like you could be hitting a path conversion error based on the shell that's being used: https://github.com/docker/cli/issues/2204
https://stackoverflow.com/questions/40213524/using-absolute-path-with-docker-run-command-not-working

This makes the fix tricky since people with self-hosted runners may be using different ENV vars/shell types on their runners.

Mannilie:
@AndrewKahr thanks for pointing me in the right direction with this!

Happy to report that I managed to find the fix to the issue. If anyone runs into this issue, the solution is quite simple. 
I noticed that on Dockers Desktop there's an option to utilize both `Linux Containers` and `Windows Containers`. 

By default, my machine was set to Linux Containers which is **_not_** which causes the following error:
image

When you switch it to `Windows Containers`:
image

The docker service should spin up without any errors:
image


Thanks again for your help @AndrewKahr , can happily close this ticket! 😄

webbertakken:
@AndrewKahr do we have a suggested approach for a PR, or recommended values for those variables on self-hosted? Perhaps we can get a quick win by streamlining this.

Related:

- https://discord.com/channels/710946343828455455/1287679642110791731
- https://github.com/game-ci/unity-builder/issues/663
- https://github.com/game-ci/unity-builder/issues/643

PR #673, opened 2024-10-03
Fix getVersionDescription() to prioritize version tags over non-version tags

OctopBP:
**Fetching Version Tags:**
The command `git tag --list --merged HEAD --sort=-creatordate` retrieves all tags that have been merged into the current branch, sorted by creation date (newest first).
We then filter these tags to keep only those that match the `grepCompatibleInputVersionRegex`, which defines valid version tags (like `1.2`).

**Handling No Version Tags:**
If no version tags are found, we fall back to using git describe with `--long` and `--always` to provide a generic description.

**Using Latest Version Tag:**
After filtering the tags, the latest valid tag is used in the git describe command, which ensures we describe the version based on the last suitable tag.

## Summary by CodeRabbit

- **New Features**
	- Enhanced version description retrieval with improved logic to prioritize valid version tags.
	- Added a warning log for better feedback when no valid version tags are found.

- **Bug Fixes**
	- Implemented a fallback mechanism to ensure version information is always available.

### DIFF
--- a/src/model/versioning.ts
+++ b/src/model/versioning.ts
@@ -207,7 +207,17 @@ export default class Versioning {
    * identifies the current commit.
    */
   static async getVersionDescription() {
-    return this.git(['describe', '--long', '--tags', '--always', 'HEAD']);
+    const versionTags = (await this.git(['tag', '--list', '--merged', 'HEAD', '--sort=-creatordate',]))
+      .split('\n')
+      .filter(tag => new RegExp(this.grepCompatibleInputVersionRegex).test(tag));
+    
+    if (versionTags.length === 0) {
+      core.warning('No valid version tags found. Using fallback description.');
+      return this.git(['describe', '--long', '--always', 'HEAD']);
+    }
+    
+    const latestVersionTag = versionTags[0];
+    return this.git(['describe', '--long', '--tags', latestVersionTag]);
   }
 
   /**
--- a/src/model/versioning.ts
+++ b/src/model/versioning.ts
@@ -209,16 +209,18 @@ export default class Versioning {
   static async getVersionDescription() {
     const versionTags = (await this.git(['tag', '--list', '--merged', 'HEAD', '--sort=-creatordate']))
       .split('\n')
-      .filter(tag => new RegExp(this.grepCompatibleInputVersionRegex).test(tag));
-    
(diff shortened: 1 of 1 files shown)


src/model/versioning.ts
```diff
@@ -207,7 +207,17 @@ export default class Versioning {
...
    */
   static async getVersionDescription() {
-    return this.git(['describe', '--long', '--tags', '--always', 'HEAD']);
+    const versionTags = (await this.git(['tag', '--list', '--merged', 'HEAD', '--sort=-creatordate',]))
+      .split('\n')
+      .filter(tag => new RegExp(this.grepCompatibleInputVersionRegex).test(tag));
+    
+    if (versionTags.length === 0) {
+      core.warning('No valid version tags found. Using fallback description.');
+      return this.git(['describe', '--long', '--always', 'HEAD']);
```
webbertakken, line 216: should this be `'--long', '--tags', '--always'`? Seems like tags is missing compared to the original version description.

webbertakken: I'm asking because this should work in all scenarios on all platforms. So Linux, MacOS, Windows either with detached head and not. The current code is battle tested in that way.

OctopBP: Thank you for these comments. You are right, all of them should be included.
Also, while checking it, I found that describe with the provided tag returns a commit count of 0, so I rewrote it a bit.

OctopBP:
I also accidentally committed `dist` files and can't revert them. I don't know if it's ok, if not, please reject it

GabLeRoux:
It is fine to commit the `dist` folder for this project. It is actually expected as per our [CONTRIBUTING guide (pull-request-prerequisites)](https://github.com/game-ci/unity-builder/blob/main/CONTRIBUTING.md#pull-request-prerequisites).

Before merging PRs, we usually make sure the `dist` folder hasn’t been tampered with, and I can confirm yours is fine. The reason you were not able to remove it is because we have a hook that handles it.

I reviewed your changes, and they seem fine. I don't have too much context, but that looks good to me.

Thanks for contributing, much appreciated!

GabLeRoux (approved):

webbertakken:
Released in [v4.3.0](https://github.com/game-ci/unity-builder/releases/tag/v4.3.0).

PR #689, opened 2025-02-07
Add initial config for devcontainers.

Egorrko:
#### Changes

-  Adding dev containers config allow contributors easily deploy project locally

repo.

- Not needed

#### Checklist

## Summary by CodeRabbit

- New Features
	- Introduced a development container configuration that simplifies setting up a Node.js and TypeScript environment with automated Yarn dependency installation.
- Documentation
	- Expanded contributor guidelines to include instructions for using development containers, ensuring a consistent and streamlined setup process for all users.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,10 @@
+{
+  "name": "Node.js & TypeScript & Yarn",
+  "image": "mcr.microsoft.com/devcontainers/typescript-node:1-22-bookworm",
+  "features": {
+    "ghcr.io/devcontainers/features/node:1": {}
+  },
+  "postCreateCommand": {
+    "Install dependencies": "yarn install"
+  }
+}
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,8 +36,30 @@ Please note that commit hooks will run automatically to perform some tasks;
 #### Windows users
 
 Make sure your editor and terminal that run the tests are set to `Powershell 7` or above with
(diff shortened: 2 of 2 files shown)


webbertakken:
If this is just for installing Node and Yarn, then we already use Volta.

Have you considered using Volta as prescribed?

Can you speak for the upsides of using devcontainer instead? What needs to be installed to use the config?

davidmfinol:
> Is it possible to test locally that the code successfully builds unity projects? I've been experimenting with [act](https://github.com/nektos/act), which allows you to run github actions locally.

Yes, you should be able to run with act.

davidmfinol (approved):
I haven't used dev containers, but I'm in support of providing the option for it.

webbertakken:
I'm not super against this approach per se.

However, I'm not sure if a much heavier stack (i.e. you now docker and an IDE extension, just to work with the repo) is helpful here. Especially since if you want the action to eventually run locally, you in addition need dind or docker service to run on the side - because it isn't always trivial to run a docker container from within another docker container. And I'm not sure how much dev containers support that kind of thing.

When looking at this PR I'm also not sure about these points:
- The devcontainer config name looks very generic and not specific to this repository
- Not everyone uses VSCode, is this really as plug and play it sounds?
- This PR introduces this new not-battle tested option as the default from the start
- Does this in fact use the correct version of yarn?
- And perhaps should it use volta, to at least keep the versioning in sync?

webbertakken:
Going to close this PR as it doesn't add enough value for the compatibility risk and maintenance.

Please vote thumbs down if you feel we need this.

Issue #716, opened 2025-07-13
document and enable caching of incremental build artifacts


js-9:
**Context**

To speed up builds, unity uses an incremental build pipeline. Caching the files produced by this pipeline has the potential to speed up the build process dramatically.

**Suggested solution**

Document where the incremental build cache is stored so people can cache that folder.
One issue is that I was not able to find the incremental cache folder on windows runners (needed for windows IL2CPP builds), so the suggested solution would be to add an option to specify the incremental build artifact directory in the action parameters (adding a new parameter for it).

**Considered alternatives**
🤷 

**Additional details**

For unchanged code, even on IL2CPP, the build takes less than 5 minutes, but I've seen 2x improvement in build time on mono build even if there are code changed.

js-9:
The PR can be considered an initial implementation to address the issue with windows paths. If someone can help me out with thouse I'd be happy to implement this in any way that people prefer, including a new actions parameter, as well as document the usage.

davidmfinol:
Instead of a PR to unity-builder, maybe this is something that should be documented on the docs at https://game.ci/docs/github/builder#caching ?

js-9:
I wasn't able to figure out the default location where the incremental build cache is stored on windows due to the limited access to the runners, which is why the PR is needed.
If/once the PR is merged I'd be happy to document the incremental build cache location for both windows and linux.

js-9:
@davidmfinol Would you mind pointing me to the repository where I can submit documentation changes?

davidmfinol:
Docs are at https://game.ci/

Each page has a link to edit it. For example, https://game.ci/docs/github/builder has a link to edit at https://github.com/game-ci/documentation/blob/main/docs/03-github/04-builder.mdx

js-9:
Thanks! https://github.com/game-ci/documentation/pull/522

frostebite:
Flagging: with v6.0.0-beta.1 (#845) delegating build/licensing logic to game-ci/cli, new features like this now belong in game-ci/cli itself rather than this wrapper. Once the migration lands, please re-file this against https://github.com/game-ci/cli if it's still wanted — happy to track it there.

PR #717, opened 2025-07-13
specify bee (incremental) build cache directory environment variable

js-9:
specify bee (incremental) build cache directory environment variable for windows docker run command and cache to Library directory
This PR can be considered as an initial proposal. I'd be happy to add an option for the unity-builder action to specify the directory in the parameters, it's just that I had a really hard time understanding how the windows runner directory layout works, and couldn't find the directory in the path that is supposed to be the default on windows.

#### Changes

- specify bee cache directory in windows docker command

#### Related Issues

- #716 - this addresses the question about the location of the incremental build artifacts on windows runners

#### Related PRs

ø

repo.

- ...

#### Checklist

### DIFF
--- a/src/model/docker.ts
+++ b/src/model/docker.ts
@@ -107,6 +107,7 @@ class Docker {
             --workdir c:${dockerWorkspacePath} \
             --rm \
             ${ImageEnvironmentFactory.getEnvVarString(parameters)} \
+            --env BEE_CACHE_DIRECTORY=c:${dockerWorkspacePath}/Library/bee_cache \
             --env GITHUB_WORKSPACE=c:${dockerWorkspacePath} \
             ${gitPrivateToken ? `--env GIT_PRIVATE_TOKEN="${gitPrivateToken}"` : ''} \
             --volume "${workspace}":"c:${dockerWorkspacePath}" \


js-9:
> I'm not understanding the goal of this PR. It looks like the bee cache directory is inside the Library. We always recommend caching the entire library folder. Why focus on this one subfolder?

Definitely! This is not quite about github actions cache as much as it is about specifying the incremental build cache location so that it can then be cached in github actions if/when somebody wants to.

davidmfinol:
Can you share some performance metrics for this?
What do build times look like with this caching enabled vs what the build times look without this?

js-9:
> Can you share some performance metrics for this? What do build times look like with this caching enabled vs what the build times look without this?

There is a dramatic speedup for mono builds. About 2x. Unfortunately, because IL2CPP build time is defined by linking, which is not cached and not parallel, the speedup for an IL2CPP build was from about 1 hour 10 minutes to about 50 minutes, which isn't negligeable but not huge. Larger projects might benefit more, though.
(I'm comparing incremental build cache-enabled build vs. builds with only Library caching enabled.)

davidmfinol:
We should be able to merge if you resolve the merge conflicts by pulling from `main` and rebuilding the `dist` folder.

davidmfinol (approved):

PR #723, opened 2025-08-02
fix: mock github checks in tests

frostebite:
Very small PR. Improving stability/fix for occasional fail on a test for coordinator/cloud runner.

Mocks a github check API which very occasionally fails.

## Summary
- polyfill fetch using undici for jest
- mock github check requests to avoid network dependency in tests

## Testing
- `cloudRunnerTests=true node node_modules/jest/bin/jest.js cloud-runner-github-checks --detectOpenHandles --forceExit --runInBand`

------
https://chatgpt.com/s/cd_688fab13de4c81918a329e14fe7b8284

## Summary by CodeRabbit

* **Tests**
  * Improved test reliability by mocking GitHub API calls in the Cloud Runner Github Checks test suite, ensuring tests run without external dependencies.
  * Enhanced Jest test setup with a polyfill for the fetch API, preventing test failures related to missing global fetch-related APIs.
  * Added a new integration test suite for Cloud Runner GitHub Checks to validate real interactions with GitHub during workflows.
* **Chores**
  * Introduced a new optional parameter to the CI workflow enabling manual execution of GitHub Checks integration tests.


cloudymax:
Seems reasonable, but since my TS is not very strong I would need to ask:
1. Why is it needed to disable certain eslint options?
2. Could mocking the github return values lead to tests passing erroneously in the event a github API changes?

frostebite:
> Seems reasonable, but since my TS is not very strong I would need to ask:
> 
> 1. Why is it needed to disable certain eslint options?
> 2. Could mocking the github return values lead to tests passing erroneously in the event a github API changes?

• Why is it needed to disable certain ESLint options? — The only disables are for camelcase where the GitHub API requires snake_case fields, so we suppress the camelCase rule around those API payload keys.
• Could mocking the GitHub return values lead to tests passing erroneously if the GitHub API changes? — Mocking keeps unit tests deterministic and offline. To catch breaking API changes, we added an optional integration test (RUN_GITHUB_INTEGRATION_TESTS) that exercises real GitHub checks, providing a safety net for API regressions.

frostebite:
> Seems reasonable, but since my TS is not very strong I would need to ask:
> 
> 1. Why is it needed to disable certain eslint options?
> 2. Could mocking the github return values lead to tests passing erroneously in the event a github API changes?

• Why is it needed to disable certain ESLint options? — The only disables are for camelcase where the GitHub API requires snake_case fields, so we suppress the camelCase rule around those API payload keys.

• Could mocking the GitHub return values lead to tests passing erroneously if the GitHub API changes? — Mocking keeps unit tests deterministic and offline. To catch breaking API changes, we added an optional integration test (RUN_GITHUB_INTEGRATION_TESTS) that exercises real GitHub checks, providing a safety net for API regressions.

@cloudymax

cloudymax:
Thanks for the explanation! - Approved 👍

Issue #737, opened 2025-09-25
Support activeBuildProfile parameter


ryo0ka:
# Context
Enhancing #674 to cover more use cases. 

Unity 6000+'s new feature "build profile" comes with multiple benefits:
1. Allows to override project/build configurations per profile
2. Allows to compile the project with profile-specific preprocessors before the build function is invoked

Current implementation by @MichaelBuhler ([5eccc32](https://github.com/MichaelBuhler/unity-builder/commit/5eccc324362c7f088693e1a44732e56666ffbc1f)) covers 1, but not 2. 

Calling `BuildProfile.SetActiveBuildProfile()` during build script will not apply preprocessors, as documented [here](https://docs.unity3d.com/2023.2/Documentation/Manual/CustomScriptingSymbols.html):
> [...] when the Editor runs in batch mode, it runs “headless”, so there is no editor loop which would cause it to recompile with the new scripting symbols. Because of this, you should not use Editor scripts to set scripting symbols within a batch mode CI server because the scripts will not be recompiled, so they will not be applied.

It'll affect projects that manage code stripping like this:
```cs
#if PROFILE_A_LOADED
using ProfileA.Specific.Module;
ProfileA.Specific.Module.DoSomething();
#endif
```

We can cover this by passing `buildProfile` to `-activeBuildProfile`. In previous discussion, we went over compatibility mitigation:
1. We continue to use `targetPlatform` to identify which Docker image to pull.
2. If `buildProfile` is given by user, we pass it to `-activeBuildProfile`; otherwise we pass `targetPlatform` to `-buildTarget`.
3. In case users made a mistake to specify a mismatching pair of `targetPlatform` and `buildProfile`, build may fail (undefined behavior).

Regarding 3, previous discussion went over the possibility of reading the profile (yaml). I personally feel like not doing it because the file content format is not documented and it's a fairly new feature.

There was another compatibility issue surrounding the default build function but @MichaelBuhler has gracefully implemented a branching logic.

# Suggested solution

Current implementation passes `buildProfile` to `-customBuildProfile`. Maybe we can simply reroute it to `-activeBuildProfile` but maybe I'm not seeing other compatibility issues?

# Considered alternatives

Running two passes of unity-build action: (1) run the action with a function that sets active build profile and (2) run the action again with another function that actually runs the build. 

# Additional details

[...]

Issue #674, opened 2024-10-03
Support activeBuildProfile parameter


Sammmte:
**Context**

When using "activeBuildProfile" parameter (in "customParameters", example: `customParameters: -activeBuildProfile Assets/Settings/BuildProfiles/WindowsDevelopment.asset`) [(for Unity 6)](https://docs.unity3d.com/6000.0/Documentation/Manual/EditorCommandLineArguments.html) builder action throws an error message:

```
Aborting batchmode due to failure:
Can't have -activeBuildProfile argument defined when using -buildTarget argument
```

Problem is "targetPlatform" is a required parameter.

**Suggested solution**

Maybe require to have either "targetPlatform" parameter OR another parameter like "buildProfile". "buildProfile" parameter would hold the same value "activeBuildProfile" requires, a relative path to the profile asset.

GabLeRoux:
Yeah Build Profiles is quite a recent feature which is indeed something that would be worth adding as a feature 👍 

[More details about the build profiles](https://docs.unity3d.com/6000.0/Documentation/Manual/build-profiles.html)

![image](https://github.com/user-attachments/assets/6b041589-0a94-4e8c-8adb-f4c1d09dd326)

OwenSignaturize:
This would be incredibly useful

GabLeRoux:
@coderabbitai read the following files and suggest an approach to add support for activeBuildProfile parameter for unity version >= 6 (6000)

dist/default-build-script/Assets/Editor/UnityBuilderAction/Builder.cs
dist/default-build-script/Assets/Editor/UnityBuilderAction/Versioning/GitException.cs
src/index.ts
src/model/platform-setup/setup-windows.ts
src/model/unity.test.ts
src/model/input.ts
dist/default-build-script/Assets/Editor/UnityBuilderAction/Versioning/VersionApplicator.cs
src/model/action.ts
dist/default-build-script/Assets/Editor/UnityBuilderAction/Versioning/Git.cs
dist/default-build-script/Assets/Editor/UnityBuilderAction/Versioning/VersionGenerator.cs
src/model/action.test.ts
src/model/unity.ts
dist/default-build-script/Assets/Editor/UnityBuilderAction/Input/ArgumentsParser.cs
dist/default-build-script/Assets/Editor/UnityBuilderAction/Reporting/StdOutReporter.cs
dist/default-build-script/Assets/Editor/UnityBuilderAction/Reporting/CompileListener.cs
src/model/input.test.ts
[...]

PR #520, opened 2023-03-04
Fix Builds Status Badge

davidmfinol:
#### Changes

- Fix Builds Status Badge

#### Checklist

- [x] Read the contribution [guide](https://github.com/game-ci/unity-builder/blob/main/CONTRIBUTING.md) and accept the
      [code](https://github.com/game-ci/unity-builder/blob/main/CODE_OF_CONDUCT.md) of conduct
- [x] Readme (updated or not needed)
- [x] Tests (added, updated or not needed)

PR #738, opened 2025-09-29
Support activeBuildProfile parameter

ryo0ka:
#### Changes

- Passing `buildProfile` to `-activeBuildProfile`

#### Related Issues

- #737 
- #674 (previous iteration)

- ✅ Ubuntu: https://github.com/ryo0ka/unity-builder/actions/runs/18084679667
- ✅Windows: https://github.com/ryo0ka/unity-builder/actions/runs/18084679670
- ❌ Mac: https://github.com/ryo0ka/unity-builder/actions/runs/18084679665

#### Checklist

- [x] Read the contribution [guide](https://github.com/game-ci/unity-builder/blob/main/CONTRIBUTING.md) and accept the [code](https://github.com/game-ci/unity-builder/blob/main/CODE_OF_CONDUCT.md) of conduct

#### Impl details

- Mac workflows failing due to an unrelated cause:

<details>

```
/Applications/Xcode_16.4.app/.../_stdio.h:318:7: error: expected identifier or '('
FILE    *fdopen(int, const char *) __DARWIN_ALIAS_STARTING(...);
...
/Applications/Unity/Hub/Editor/2022.3.13f1/Unity.app/Contents/il2cpp/external/zlib/zutil.h:147:33: note: expanded from macro 'fdopen'
#define fdopen(fd,mode) NULL /* No fdopen() */
```

AI analysis (not validated):
> This happens because zlib's zutil.h defines fdopen as a macro, which conflicts with the standard library's fdopen declaration in the MacOS 15 SDK. This is a known compatibility issue when building Unity projects on newer MacOS SDKs and Xcode releases.

> Workaround: Force Unity to use an older MacOS SDK (such as MacOS 13 or 14) and an Xcode version compatible with Unity 2022.3 LTS. Unity 2022.3.x may not yet support Xcode 16.4 and MacOS 15 SDK.

Since this is out of scope, I didn't investigate further.

</details>

- Windows/Mac workflows don't include Unity 6000 + WebGL matrix, so the new path is only tested on Ubuntu workflows. Windows/Mac workflows only ensure backward compatibility with older Unity versions.
- `-activeTarget` compatibility -- New implementation omits `-activeTarget` from Unity CLI args if `buildProfile` is passed to game-ci action. This aligns with Unity's behavior and is discussed in the linked issue. If user passed `buildProfile` to a project of older versions (2023-) by mistake, the build will likely fail. 
- Previous iteration (#674) introduced a custom Unity CLI argument `-customBuildProfile` which is picked up in `default-build-script` to set the active build profile during build process. I assume that this is working around the `-activeTarget` compatibility issue. New implementation (with `-activeBuildProfile` official CLI argument) loads the build profile before the build script runs, so the custom argument is not used anymore. I've left it in the project for backward compatibility.
- New implementation adds support for profile-specific define symbols, which is an official feature that allows to customize script compilation. I've added a symbol `BUILD_PROFILE_LOADED` to the existing profile in `test-project`, which is properly loaded to the build script in `default-build-script`, otherwise the integration test will fail.

### DIFF
--- a/dist/default-build-script/Assets/Editor/UnityBuilderAction/Builder.cs
+++ b/dist/default-build-script/Assets/Editor/UnityBuilderAction/Builder.cs
@@ -56,7 +56,10 @@ namespace UnityBuilderAction
       // of either `UnityEditor.BuildPlayerOptions` or `UnityEditor.BuildPlayerWithProfileOptions`
       dynamic buildPlayerOptions;
 
-      if (options.TryGetValue("activeBuildProfile", out var buildProfilePath) && !string.IsNullOrEmpty(buildProfilePath)) {
+      if (options.TryGetValue("activeBuildProfile", out var buildProfilePath)) {
+        if (string.IsNullOrEmpty(buildProfilePath)) {
+          throw new Exception("`-activeBuildProfile` is set but with an empty value; this shouldn't happen");
+        }
 
 #if UNITY_6000_0_OR_NEWER
         // Load build profile from Assets folder


dist/default-build-script/Assets/Editor/UnityBuilderAction/Builder.cs
```diff
@@ -56,27 +56,37 @@ public static void BuildProject()
       // of either `UnityEditor.BuildPlayerOptions` or `UnityEditor.BuildPlayerWithProfileOptions`
       dynamic buildPlayerOptions;
 
-      if (options["customBuildProfile"] != "") {
+      if (options.TryGetValue("activeBuildProfile", out var buildProfilePath) && !string.IsNullOrEmpty(buildProfilePath)) {
```
webbertakken: Shall we go with a descriptive error, in case of null or empty?

webbertakken (approved):

dist/platforms/ubuntu/steps/build.sh
```diff
@@ -125,10 +125,11 @@ unity-editor \
   $( [ "${MANUAL_EXIT}" == "true" ] || echo "-quit" ) \
   -customBuildName "$BUILD_NAME" \
   -projectPath "$UNITY_PROJECT_PATH" \
-  -buildTarget "$BUILD_TARGET" \
+  $( [ -z "$BUILD_PROFILE" ] && echo "-buildTarget $BUILD_TARGET" ) \
   -customBuildTarget "$BUILD_TARGET" \
   -customBuildPath "$CUSTOM_BUILD_PATH" \
   -customBuildProfile "$BUILD_PROFILE" \
```
timcassell, line 131: It looks like `customBuildProfile` is no longer needed here (and other platforms).

ryo0ka: Yes, it's not needed/used anymore in `default-build-script`, but user could be picking it up in their own build script via `-executeMethod`. Example code of using the old argument exists in the previous iteration, which is public, so I decided to leave it there.

dist/default-build-script/Assets/Editor/UnityBuilderAction/Builder.cs
ryo0ka: commited a fix (mac/windows workflows cancelled)

webbertakken:
@ryo0ka can you confirm this works? Is it in a mergeable state?

ryo0ka:
@webbertakken yes, I did run Ubuntu workflows on the latest commit and didn't see an error

webbertakken:
Released in [v4.6.0](https://github.com/game-ci/unity-builder/releases/tag/v4.6.0)

somnisomni:
Hello, I think this change seems break my automatic build pipeline. Build was successful before this change got merged.

**Environment**:
- Running on GitHub Actions self-hosted runner (Windows)
- Unity 6000.0.42f1

**Workflow job definition**:
```
- name: Build
        uses: game-ci/unity-builder@main
        env:
        with:
          versioning: None  # We have dedicated versioning pipeline in the source code
          runAsHostUser: true
          buildName: '[REDACTED]'
          buildProfile: 'Assets/06_Settings/Build Profiles/Windows Dev.asset'
```

**Command line arguments according to actions log**:
```
COMMAND LINE ARGUMENTS:
C:\UnityEditor\6000.0.42f1\Editor\Unity.exe
-quit
-batchmode
-nographics
-silent-crashes
-customBuildName
[REDACTED]
-projectPath
c:/github/workspace\.
-executeMethod
UnityBuilderAction.Builder.BuildProject
-customBuildTarget
StandaloneWindows64
-customBuildPath
c:/github/workspace\build/StandaloneWindows64\[REDACTED].exe
-customBuildProfile
Assets/06_Settings/Build Profiles/Windows Dev.asset
-buildVersion
none
-androidVersionCode
0
-androidKeystorePass
-androidKeyaliasName
-androidKeyaliasPass
-androidTargetSdkVersion
-androidExportType
androidPackage
-androidSymbolType
none
-logfile
-
-activeBuildProfile
Assets/06_Settings/Build
Profiles/Windows
Dev.asset
```

**Key error message**:
```
Unable to open the specified build profile 'Assets/06_Settings/Build'

Aborting batchmode due to failure:
Unable to open the specified build profile 'Assets/06_Settings/Build'
```

In command line arguments, the value of `activeBuildProfile` got 'splitted' for each whitespaces, making each splitted segments recognized as individual parameters. I think this is the key problem.
And yes, I can move the build profile file to somewhere without whitespaces in the path, but I don't think this is not an expected behavior so I'm reporting this.

jonTabi:
I ran into the bug. Unity's default folder path seems to be "Build Profiles" annoyingly. 

Taking a quick look, my hunch is that the fix is to wrap the arguments in build.ps1, build.sh with the quotes as like the other variables, but I don't know much about shell/ps so someone else should confirm.

E.g. 
$unityArgs += @("-activeBuildProfile", $Env:BUILD_PROFILE)
should be
$unityArgs += @("-activeBuildProfile", "`"$Env:BUILD_PROFILE`"")

"-buildTarget $BUILD_TARGET" is missing proper quotes around $BUILD_TARGET

webbertakken:
Ah yes, taking PRs for the quoting/escaping issue!

webbertakken:
@ryo0ka would you mind doing the follow-up PR?

ryo0ka:
@webbertakken this issue is because of this:
> Windows/Mac workflows don't include Unity 6000 + WebGL matrix, so the new path is only tested on Ubuntu workflows. Windows/Mac workflows only ensure backward compatibility with older Unity versions.

I'll add Unity 6000+ paths to Windows/MacOS workflows which are currently missing. That should expose the quoting/escaping issue & I'll fix that too.

Regarding #741, it's due to me failing to update the document. @davidmfinol has dealt with it in #520.

mob-sakai:
The related issue: https://github.com/game-ci/unity-builder/issues/742

mob-sakai:
If the profile does not contain 'BUILD_PROFILE_LOADED' as 'Scripting Defines', the build will fail.

PR #764, opened 2025-12-05
Add support for parallel linking in IL2CPP builds

js-9:
#### Changes
This adds support for and enables by default parallel linking for IL2CPP builds.
IL2CPP builds spend a lot of time in the linking stage, since the compilation step is already parallel. This updates the build to make unity use parallel linking, which can significantly speed up builds, especially if the library and incremental build cache is used, in which cases linking is where most of the build time is spent.
* Introduced 'enableParallelLinking' input in action.yml to control parallel linking.
* Updated Builder.cs to apply parallel linking settings based on the new input.
* Modified build scripts for macOS, Ubuntu, and Windows to include the parallel linking option.

I've been using this in our private repo for a bit and it seems to work, although ios, osX and Android builds are untested.

#### TODOs
If people are interested in this change I'll update documentation and README as well.
- [ ] Update README
- [ ] Update Documentation

#### Checklist

- [ ] Docs (If new inputs or outputs have been added or changes to behavior that should be documented. Please make a PR

## Summary by CodeRabbit

* **New Features**
  * Added an optional enableParallelLinking input to the Unity - Builder Action (enabled by default) to allow parallel IL2CPP linking for faster builds.
  * The setting is exposed to build environments via an ENABLE_PARALLEL_LINKING variable so build tools on all supported platforms (including macOS runners) can act on it.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -43,6 +43,10 @@ inputs:
     required: false
     default: ''
     description: 'Launches unity without specifying `-nographics`.'
+  enableParallelLinking:
+    required: false
+    default: 'true'
+    description: 'Enable parallel linking for IL2CPP builds to speed up linking phase.'
   customParameters:
     required: false
     default: ''
--- a/dist/default-build-script/Assets/Editor/UnityBuilderAction/Builder.cs
+++ b/dist/default-build-script/Assets/Editor/UnityBuilderAction/Builder.cs
@@ -116,6 +116,12 @@ namespace UnityBuilderAction
           AndroidSettings.Apply(options);
         }
 
+        // Enable parallel linking for IL2CPP builds
+        if (options.TryGetValue("enableParallelLinking", out var enableParallelLinking) && 
+            enableParallelLinking != "false") {
+          SetParallelLinking(buildTarget);
+        }
+
       }
 
       // Perform build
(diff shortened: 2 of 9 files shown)


js-9:
> **Actionable comments posted: 0**
> 🧹 Nitpick comments (1)
> 
> > src/model/input.ts (1)> `146-150`: **Boolean getter works; consider whether default should be `true` everywhere**
> > > The getter follows the existing pattern for `manualExit` / `enableGpu` and will correctly return `true` when the action input is `'true'`. In GitHub Actions runs, the `action.yml` default `'true'` means this flag is enabled by default, but in non‑GitHub contexts (e.g. CLI/CloudRunner without explicit input) the `?? false` fallback makes the default effectively `false`.
> > > If you want “parallel linking on by default” consistently across all entrypoints, consider defaulting the fallback to `'true'` instead:
> > > ```ts
> > > const input = Input.getInput('enableParallelLinking') ?? 'true';
> > > return input === 'true';
> > > ```
> 
> 📜 Review details

The nitpick comment here has been addressed in onibi-gg/unity-builder@ff72feceddfd9837c62155de0171ccfcaa716657 .

timcassell:
Is that flag documented somewhere? I couldn't find it from a google search.

It would be nice to add it to unity-test-runner also to speed up standalone builds.

js-9:
> Is that flag documented somewhere? I couldn't find it from a google search.
> 
> It would be nice to add it to unity-test-runner also to speed up standalone builds.

This only has effect with IL2CPP since mono builds do not require a linking stage. I'd be surprised if the test runner used IL2CPP, I can't see much benefit to it except in very limited situations where code would have different results on mono vs. IL2CPP.
Anyways, the flags are documented here: https://docs.unity3d.com/6000.2/Documentation/Manual/handling-il2cpp-additional-args.html

timcassell:
I use it to test il2cpp builds for my library. The build part takes up half the test time, so any speedup would be appreciated. 🙂

js-9:
> I use it to test il2cpp builds for my library. The build part takes up half the test time, so any speedup would be appreciated. 🙂

Ah. Does the test-runner have a separate repo and build script? This same technique should work on that too, I'd guess.

js-9:
BTW I need to update the patch, since this looks at the targetplatform to set platform-specific linker flags, but I need to change it to look at the build platform since that is where the compiler differences are.

EDIT: this is done

timcassell:
https://github.com/game-ci/unity-test-runner/blob/main/dist/platforms/ubuntu/run_tests.sh

js-9:
I'll see what I can do.

js-9:
For visibility: https://github.com/game-ci/unity-test-runner/pull/300#issuecomment-3619634533

frostebite:
Thanks for the PR! This is superseded by the thin-wrapper migration (currently open as the "Thin wrapper: invoke game-ci/cli as a subprocess" PR in this repo): it patches `src/model/` and the bundled `default-build-script/Assets/Editor/UnityBuilderAction/Builder.cs`, both of which go away once that migration lands — build orchestration and the injected Editor script move into game-ci/cli itself. It's also showing merge conflicts against current `main`.

Parallel IL2CPP linking is a legitimate feature request, but it now needs to be implemented in game-ci/cli directly. Closing here — please open the equivalent request against https://github.com/game-ci/cli so it can be tracked there.

PR #819, opened 2026-03-09
Orchestrator Extraction — Remove Orchestrator Code, Add Plugin Loader

frostebite:
## Orchestrator Extraction — Remove Orchestrator Code, Add Plugin Loader

Extracts the orchestrator subsystem from unity-builder into the standalone [`game-ci/orchestrator`](https://github.com/game-ci/orchestrator) repository. Unity-builder becomes a lean GitHub Action; all orchestrator, CLI, and cloud provider features live in the standalone repo.

> **Standalone repo:** https://github.com/game-ci/orchestrator (public)
> **Documentation:** https://github.com/game-ci/documentation/pull/541
> **Closes PRs:** #777, #778, #783, #784, #786, #787, #790, #791, #798, #799, #804, #806, #808, #809, #813, #814, #815, #816

---

### Why

| Problem | Impact |
|---------|--------|
| 9 heavy AWS/K8s npm deps bundled into every action run | ~15MB bundle bloat for users who never use orchestrator |
| 148+ production + test files in `src/model/orchestrator/` | Contributor confusion — orchestrator internals dwarf the core action |
| Orchestrator and core action share one release cadence | Cannot ship orchestrator fixes without touching the action |
| `BuildParameters` is a god object mixing both concerns | Tight coupling makes both harder to evolve |

---

### What Changes for Users

The orchestrator is now a **separate package** that lives in its own repository ([`game-ci/orchestrator`](https://github.com/game-ci/orchestrator)). What this means depends on your build strategy:

#### Local builds (`providerStrategy: local`) — No change required

If you use the default local Docker builds, **nothing changes**. Unity-builder works standalone with no extra install step:

```yaml
steps:
  - uses: actions/checkout@v4

  - uses: game-ci/unity-builder@v4
    with:
      targetPlatform: StandaloneLinux64
```

#### Cloud builds (`providerStrategy: aws`, `k8s`, etc.) — Install the orchestrator first

For cloud provider strategies, you now need to install the orchestrator **before** running unity-builder. The install adds the `game-ci` CLI and the `@game-ci/orchestrator` package so that unity-builder's plugin loader can detect it automatically.

**GitHub Actions example:**

```yaml
steps:
  - uses: actions/checkout@v4

  - name: Install Game CI Orchestrator
    run: curl -fsSL https://raw.githubusercontent.com/game-ci/orchestrator/main/install.sh | sh

  - uses: game-ci/unity-builder@v4
    with:
      providerStrategy: aws
[...]

PR #823, opened 2026-03-27
Orchestrator aws caching fix - gracefully keep cache valid if OOM crash occurs

frostebite:
## Summary

- Remove empty placeholder tar creation (`tar --files-from /dev/null`) in `build-automation-workflow.ts` that was creating ~10KB stub files before `remote-cli-post-build` runs
- Add safety-net size check in `aws-s3-upload-cache` hook to delete tar files < 1KB before S3 upload
- Keep `mkdir -p` for cache directories so hooks don't error on missing dirs

## Problem

When a build container is OOM-killed before `remote-cli-post-build` can create real Library/build cache tars, only the empty placeholder tars (~10KB) survive. These get uploaded to S3 by the upload hooks. On the next build, `aws-s3-pull-cache` downloads and extracts these empty tars, giving Unity an empty Library folder — zero caching benefit, full reimport every time.

This was reported by a user with a 20GB project running with `containerMemory: 4096` — every build took 5+ hours because the Library cache was never populated.

## Test plan

- [ ] Verify existing e2e caching test (`orchestrator-end2end-caching.test.ts`) still passes — it expects real Library tars from `PushToCache()`, not the removed placeholders
- [ ] Run a build with `providerStrategy: aws` and confirm real cache tars are created by `remote-cli-post-build` and uploaded to S3
- [ ] Simulate OOM (low memory) and confirm no empty tars are uploaded to S3

## Summary by CodeRabbit

* **Bug Fixes**
  * Cache upload process now automatically removes small temporary cache files under 1KB before storing to remote buckets, eliminating unnecessary clutter
  * Enhanced cache preparation by eliminating empty placeholder archive file generation, resulting in cleaner storage and more efficient cache management

### DIFF
--- a/src/model/orchestrator/services/hooks/container-hook-service.ts
+++ b/src/model/orchestrator/services/hooks/container-hook-service.ts
@@ -155,6 +155,10 @@ export class ContainerHookService {
       fi
       ENDPOINT_ARGS=""
       if [ -n "$AWS_S3_ENDPOINT" ]; then ENDPOINT_ARGS="--endpoint-url $AWS_S3_ENDPOINT"; fi
+      # Skip uploading empty or near-empty tar files (< 1KB) — these are leftover
+      # stubs with no real cache data and would poison the cache for the next build.
+      find /data/cache/$CACHE_KEY/lfs -name "*.tar*" -size -1k -delete 2>/dev/null || true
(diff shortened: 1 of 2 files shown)


frostebite:
Closing — this fix has been ported directly to the standalone orchestrator repo (game-ci/orchestrator@374ead0). Since PR #819 removes all orchestrator code from unity-builder, this PR is no longer needed here.

PR #834, opened 2026-05-06
feat: add linux64RemoveExecutableExtension (default: keep .x86_64)

### LINKED ISSUE #722: Allow specifying the use of a file extension for StandaloneLinux64
**Context**

Currently, when building a game for StandaloneLinux64, this action will produce an executable without a file extension. This is fine, but is a mismatch for what the unity editor will do by default (produces an executable with the extension `.x86_64`), may cause some packages to build incorrectly (<https://github.com/ValveSoftware/unity-xr-plugin/issues/139>), and technically violates unity's (only recently?) documented [build path requirements](<https://docs.unity3d.com/6000.1/Documentation/Manual/build-path-requirements.html>).

**Considered solutions**

Keeping in mind that other people may be relying on this lack of a file extension, we cannot simply change the default without possibly causing breakage.

The least configurable option would be to add a parameter `bikeshedLinuxUseFileExtension: true` with a default to `false` which causes the file extension to be `.x86_64` when building for StandaloneLinux64. (or as appropriate for StandaloneLinux32/older versions of unity.) In the future, it may be viable to flip the default.

Another option would be to add a parameter `bikeshedLinuxFileExtension: string` which sets the file extension, allowing the user to configure it to anything.

**Considered alternatives**

Fixing <https://github.com/ValveSoftware/unity-xr-plugin/issues/139> upstream or downstream.

frostebite:
## Summary

Adds configurable `linux64RemoveExecutableExtension` parameter for StandaloneLinux64 builds.

**Default: `false`** — keeps the `.x86_64` extension, matching Unity's [build path requirements](https://docs.unity3d.com/6000.1/Documentation/Manual/build-path-requirements.html). Set to `true` to restore the extensionless behavior from v4.

This is a **breaking change** from v4 behavior (where the extension was removed by default). Bundled with the v5 major version bump.

## Context

Rebased from @kitlith's #726 onto current main — the original branch had 120+ file conflicts from the orchestrator extraction. Default flipped from `true` to `false` for v5 as discussed in #726.

Closes #722
Supersedes #726

## Test plan

- [x] 97 tests pass (build-parameters + input tests)
- [x] New test cases cover both `true` and `false` values for Linux64
- [x] Existing Android/Windows extension tests unaffected

## Summary by CodeRabbit

* **New Features**
  * Added optional input parameter to control removal of the .x86_64 executable extension when building Linux 64-bit targets. Default is false, preserving the extension. This provides enhanced flexibility in managing executable naming conventions for StandaloneLinux64 builds across different deployment environments.


frostebite:
Combined into a single PR.

PR #836, opened 2026-05-06
chore: remove legacy CLI bootstrap (commander-ts, reflect-metadata)

frostebite:
## Summary

Removes the legacy CLI bootstrap that was left behind after the orchestrator extraction (#819).

**Removed:**
- `InitCliMode()` / `RunCli()` from `Cli` class
- `@CliFunction` decorator and `CliFunctionsRepository`
- `cli-functions-repository.ts` file
- 3 dependencies: `commander-ts`, `reflect-metadata`, `commander`

**Kept:**
- `Cli.options`, `Cli.isCliMode`, `Cli.query()` — still used by the orchestrator plugin

The only registered CLI mode was `print-input` (debug command). All real CLI functionality lives in the orchestrator repo now.

## Test plan

- [x] 341 tests pass (23 suites)
- [x] 3 fewer production dependencies

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Release Notes

* **Refactor**
  * Removed CLI-mode routing functionality
  * Removed CLI functions repository and decorator
  * Consolidated import structure

* **Chores**
  * Removed unused dependencies (commander, commander-ts, reflect-metadata)

PR #835, opened 2026-05-06
chore: bump production dependencies (minor/patch)

frostebite:
## Summary

Safe minor/patch dependency bumps:

| Package | From | To |
|---|---|---|
| `@actions/cache` | ^4.0.0 | ^4.1.0 |
| `@actions/github` | ^6.0.0 | ^6.0.1 |
| `commander` | ^9.0.0 | ^9.5.0 |
| `nanoid` | ^3.3.1 | ^3.3.12 |
| `reflect-metadata` | ^0.1.13 | ^0.2.2 |
| `semver` | ^7.5.2 | ^7.7.4 |
| `yaml` | ^2.2.2 | ^2.8.4 |

Major bumps (`@actions/core` 3.x, `@actions/exec` 3.x, `nanoid` 5.x ESM-only, `@actions/github` 9.x) deferred — they require API changes.

Note: `reflect-metadata` and `commander-ts` are still needed — the legacy CLI bootstrap in `src/model/cli/cli.ts` uses `@CliFunction` decorators. Can be removed when the CLI path is fully delegated to the orchestrator.

Supersedes the Snyk PRs #693, #707, #831.

## Test plan

- [x] 341 tests pass (23 suites)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Updated multiple project dependencies to newer versions, including utility libraries and GitHub Actions integration tools, to enhance stability and ensure better compatibility.

PR #837, opened 2026-05-06
chore: v5 prep — dep bumps, linux64 extension, legacy CLI removal, Cli→PluginOptions rename

### LINKED ISSUE #722: Allow specifying the use of a file extension for StandaloneLinux64

frostebite:
## Summary

Combined PR for v5 preparation. Replaces #834, #835, #836.

### 1. linux64RemoveExecutableExtension (from @kitlith's #726)
- New `linux64RemoveExecutableExtension` parameter (default: `false`)
- Keeps `.x86_64` extension by default, matching Unity's native behavior
- Set to `true` to restore v4's extensionless behavior
- **Breaking change** from v4 — appropriate for v5 bump
- Closes #722

### 2. Dependency bumps (minor/patch)
| Package | From | To |
|---|---|---|
| `@actions/cache` | ^4.0.0 | ^4.1.0 |
| `@actions/github` | ^6.0.0 | ^6.0.1 |
| `nanoid` | ^3.3.1 | ^3.3.12 |
| `reflect-metadata` | ^0.1.13 | ^0.2.2 |
| `semver` | ^7.5.2 | ^7.7.4 |
| `yaml` | ^2.2.2 | ^2.8.4 |

### 3. Legacy CLI removal
- Removed `InitCliMode()`, `RunCli()`, `@CliFunction` decorator, `CliFunctionsRepository`
- Dropped 3 deps: `commander-ts`, `reflect-metadata`, `commander`
- Only registered CLI mode was `print-input` (unused debug command)

### 4. Cli → PluginOptions rename
- `Cli` class renamed to `PluginOptions` (it's a plugin options bridge, not a CLI)
- Backwards-compatible `Cli` alias exported for orchestrator compatibility
- Moved from `src/model/cli/cli.ts` to `src/model/plugin-options.ts`
- `isCliMode` → `isPluginMode`

## Test plan
- [x] 346 tests pass (23 suites)
- [x] 3 fewer production dependencies

## Summary by CodeRabbit

* **New Features**
  * Added a configuration input to control whether Standalone Linux64 builds include the .x86_64 executable extension (defaults to false).

* **Bug Fixes**
  * Build output now honors the Linux64 extension setting across build scenarios.

* **Tests**
  * Expanded tests to cover the new Linux64 extension behavior and input parsing.

* **Chores**
  * Updated dependencies and CI test runner/workflow commands.


src/model/plugin-options.ts
```diff
@@ -0,0 +1,36 @@
...
+ * from the options map instead of @actions/core.getInput().
+ */
+export class PluginOptions {
+  public static options: Record<string, any> | undefined;
+
+  static get isPluginMode() {
+    return (
+      PluginOptions.options !== undefined &&
+      PluginOptions.options.mode !== undefined &&
+      PluginOptions.options.mode !== ''
```
webbertakken: You'd normally not use undefined explicitly like this.

We should probably move to `PluginOptions.options?.mode? !== ''` if this is a much repeated pattern

webbertakken (approved):

src/model/plugin-options.ts
frostebite: Done in 527fd6a: simplified this to Boolean(PluginOptions.options?.mode). Verified with yarn tsc --noEmit and yarn vitest run src/model/input.test.ts src/model/build-parameters.test.ts.

PR #838, opened 2026-05-07
fix: license server toolset input + runAsHostUser HOME/USER (closes #739)

### LINKED ISSUE #739: BuildTarget is ignored on linux
**Bug description**

BuildTarget is ignored on linux building machine, The editor will start in Linux standalone no matter what
**How to reproduce**

on Windows this script

      - name: Build Android Package
        id: build-android-package
        if: ${{ github.event.inputs.debug == 'false' || github.event_name != 'workflow_dispatch' }}
        uses: game-ci/unity-builder@v4.3.0
        env:
          BUILD_APP_BUNDLE: ${{ env.BUILD_TYPE == 'aab' && 'true' || 'false' }}
        with:
          targetPlatform: Android
          customImage: ${{ env.DOCKER_IMAGE }}
          buildMethod: ${{ env.buildMethod }}
          projectPath: ${{ env.WORKSPACE_PATH }}/lm_client # Updated project path
          version: "1.0.0"
          versioning: 'Custom'
          customParameters: -nographics -silent-crashes -batchmode -quit -buildTarget Android -aabbundleversion ${{ needs.validate-build.outputs.BUNDLE_VERSION }} -keystorepass ${{ steps.keystore-params.outputs.KEYSTORE_PASS }} -keyalias ${{ steps.keystore-params.outputs.KEY_ALIAS }} -keypass ${{ steps.keystore-params.outputs.KEY_ALIAS_PASS }} -keystorebase64 ${{secrets.KEYSTORE_BASE64}} -isBM yes -adddefines ${{ github.event.inputs.extra-defines }} 

started unity in Android mode, but on linux, regardless what I use, the editor always start in linux standalone mode.

I tried using 4.5.0 and 4.6.0 with the new build profile option and no change. The command line shows it should run in android mode

> COMMAND LINE ARGUMENTS: /opt/unity/Editor/Unity -batchmode -logfile /dev/stdout -quit -customBuildName Android -projectPath /github/workspace/test-android/lm_client -customBuildTarget Android -customBuildPath /github/workspace/build/Android/Android.apk -customBuildProfile Assets/Settings/Android-Profiling.asset -activeBuildProfile Assets/Settings/Android-Profiling.asset -executeMethod Lm.Editor.BuildSupport.BuildAndroidApk -buildVersion 1.0.0 -androidVersionCode 1000000 -androidKeystoreName -androidKeystorePass -androidKeyaliasName -androidKeyaliasPass -androidTargetSdkVersion -androidExportType androidPackage -androidSymbolType none -nographics -silent-crashes -batchmode -quit -aabbundleversion 410 -keystorepass -keyalias -keypass -keystorebase64 *** -isBM yes -adddefines Successfully changed project path to: /github/workspace/test-android/lm_client

but there is nothing to do, it will always start without the UNITY_ANDROID define (and with the UNITY_STANDALONE_LINUX define enabled). Again this worked on windows, but is not working on linux no matter what. the image I use are:

unityci/editor:windows-6000.0.26f1-android-3.1
unityci/editor:ubuntu-6000.2.6f1-android-3.1.0

so as you can see the unity version is a factor too.

frostebite:
## What this is

Two opt-in, backwards-compatible fixes for the issues users hit on Linux Docker builds with a Unity floating license server, surfaced in #739.

The original symptom in #739 was `-buildTarget Android` being silently ignored on Linux. Investigation in the issue thread revealed it's actually a licensing problem: when the Licensing Client can't resolve the right entitlement, Unity falls through to a license that lacks the requested platform support, and the editor produces a Linux Standalone artifact without raising an error. Two distinct root causes surfaced — addressed below.

### 1. `unityLicensingToolset` input

Floating-license servers can host multiple toolsets. Without a `toolset` field in `services-config.json`, the Licensing Client relies on the server-side default toolset; if the server doesn't have one set (or the wrong one is set) it returns `404 / Found 0 entitlement groups`, and Unity falls through.

This PR adds an optional `unityLicensingToolset` input, threaded through `Input` → `BuildParameters` → `SetupShared`, where it's injected into the rendered `services-config.json` only when set. The Licensing Client then requests entitlements from the named toolset.

### 2. `runAsHostUser` license-client malfunction

When `runAsHostUser: true`, `entrypoint.sh` invoked `su $USERNAME -c "..."` without explicit `HOME`/`USER`. The host user inherited root's environment (`HOME=/root`, `USER` unset), so the Unity Licensing Client — which writes to `~/.config/unity3d` — could not resolve a writable home directory. This matches OP's observation in the issue thread that dropping `runAsHostUser: true` resolved their licensing failure even after they reverted the toolset workaround.

Fix: set `HOME`/`USER`/`LOGNAME` explicitly on the inner shell command, while preserving the rest of root's env via `su -p`.

## Are both opt-in? Are both backwards compatible?

| Fix | Opt-in | Backwards compatible | Notes |
|---|---|---|---|
| `unityLicensingToolset` input | Yes — default empty, field omitted from `services-config.json` when unset | Yes — byte-for-byte identical config when unset | None |
| `runAsHostUser` HOME/USER fix | Yes — entirely inside the `if RUN_AS_HOST_USER == "true"` branch | **Mostly.** Only changes env for users *currently* using `runAsHostUser: true`. Anyone whose pipeline silently relies on the broken `HOME=/root` would see paths shift to `/home/$USER`. Almost certainly nobody does — but it is technically a behavior change for that opt-in branch. |

Users not using a license server and not using `runAsHostUser` see zero change.

## What I'm uncertain about

- **No reproduction.** I can't reproduce either issue locally — none of the Linux license-server flows are exercisable from a maintainer machine without a live floating-license server. The fixes are derived from the logs and behavior reported in #739.
- **The `runAsHostUser` fix is plausible but not confirmed.** OP (sebas77) hedged on the cause ("almost definitively") and moved off game-ci before verifying. The `/home/UNKNOWN` log line plus inspection of `entrypoint.sh` makes this the strongest hypothesis — but it would be ideal if a reviewer with access to a license server validated before merge. Happy to split it into a follow-up PR if you'd prefer to land the toolset fix alone first.

## What I considered and dropped

- A general "merge arbitrary JSON into `services-config.json`" input. Larger API surface, conflicts with existing `unityLicensingServer`, footgun in YAML. Single named input is narrower and safer; if more fields become useful, add them one at a time.
- Fail-fast on Licensing Client errors in the Unity log. The pattern that fired here (`com.unity.editor.headless not found`) isn't always fatal — Unity probes entitlements during startup and may succeed on a later one — and log formats drift across Unity 5/2019/2022/6000/7000. False-firing would turn currently-green builds red. Worth its own investigation.

## Companion change

Mirrored across the orchestrator CLI surface: game-ci/orchestrator#TBD (linked once opened). The `entrypoint.sh` HOME/USER fix flows through automatically because orchestrator copies it from `unity-builder/dist/platforms/ubuntu/`.

## Test plan

- [x] `yarn typecheck` clean
- [x] `yarn lint` clean (only pre-existing warnings)
- [x] `yarn test:ci` — 347 passed, 2 skipped (no failures)
- [x] `yarn build` regenerates `dist/index.js` containing the new field
- [ ] Verify on a real floating-license server that `unityLicensingToolset` is honored (cannot do locally — need a reviewer with access)
- [ ] Verify on a real Linux runner with `runAsHostUser: true` that licensing succeeds (cannot do locally — need a reviewer with access)

Closes #739.

## Release Notes

* **New Features**
  * Added optional `unityLicensingToolset` input for configuring the named toolset identifier in Unity floating-license server scenarios.

* **Documentation**
  * Updated `StandaloneLinux64` input description to explicitly note restoration of extensionless behavior from v4.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -168,6 +168,14 @@ inputs:
     default: ''
     required: false
     description: 'The Unity licensing server address to use for activating Unity.'
+  unityLicensingToolset:
+    default: ''
+    required: false
+    description:
+      'Optional toolset identifier for Unity floating-license servers that host multiple toolsets (e.g.
+      "LicenseServer_1234567890_3"). When set, written to services-config.json so the Licensing Client
+      requests entitlements from the named toolset instead of relying on the server-side default. Leave
+      empty to preserve previous behavior.'
   dockerWorkspacePath:
     default: '/github/workspace'
     required: false
@@ -181,8 +189,7 @@ inputs:
   linux64RemoveExecutableExtension:
     default: 'false'
     required: false
-    description:
-      'When building for StandaloneLinux64, remove the default file extension of `.x86_64`. Set to true to restore the extensionless behavior from v4.'
+    description: 'When building for StandaloneLinux64, remove the default file extension of `.x86_64`. Set to true to restore the extensionless behavior from v4.'
 
 outputs:
   volume:
--- a/dist/licenses.txt
+++ b/dist/licenses.txt
@@ -870,51 +870,6 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 
 
-commander
-MIT
-(The MIT License)
-
-Copyright (c) 2011 TJ Holowaychuk <tj@vision-media.ca>
-
-Permission is hereby granted, free of charge, to any person obtaining
-a copy of this software and associated documentation files (the
-'Software'), to deal in the Software without restriction, including
-without limitation the rights to use, copy, modify, merge, publish,
-distribute, sublicense, and/or sell copies of the Software, and to
-permit persons to whom the Software is furnished to do so, subject to
-the following conditions:
-
-The above copyright notice and this permission notice shall be
-included in all copies or substantial portions of the Software.
-
-THE SOFTWARE IS PROVIDED 'AS IS', WITHOUT WARRANTY OF ANY KIND,
-EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
-MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
-IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
-CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
-TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
-SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
-
-
-commander-ts
-ISC
-Copyright 2018 Matt Strom <matt@mattstrom.com>
-
-Permission to use, copy, modify, and/or distribute this software
-for any purpose with or without fee is hereby granted, provided
-that the above copyright notice and this permission notice appear
-in all copies.
-
-THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL
-WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED
-WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL
-THE AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR
-CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM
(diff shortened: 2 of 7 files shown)


frostebite:
unity-builder is now a thin wrapper around [game-ci/cli](https://github.com/game-ci/cli) ([#844](https://github.com/game-ci/unity-builder/pull/844)) and no longer ships its own `dist/platforms` scripts - it downloads and shells out to the CLI's own copy at runtime. So the two fixes here needed to land there instead:

- The `runAsHostUser` HOME/USER/LOGNAME fix: ported as [game-ci/cli#197](https://github.com/game-ci/cli/pull/197), released as `game-ci/cli` v0.1.36.
- The `unityLicensingToolset` input: already present in `game-ci/cli` (added independently, threaded through to `services-config.json` the same way this PR does it) - nothing further needed there.

Closing this in favor of the CLI-side port, but thanks for the thorough root-causing here (and for flagging your own uncertainty about the `runAsHostUser` behavior change honestly - it held up).

PR #849, opened 2026-08-28
ci: auto-retry test runs when every failed job matches a known-transient pattern

frostebite:
## Summary
Adds a narrow, classified auto-retry for this repo's own build-tests-* CI, closing the gap the in-process Unity license retry (`activate.sh`/`build.sh`, up to 4 attempts with exponential backoff) can't cover: that retry reuses the same runner, so it's powerless against a runner-specific issue (e.g. a stuck Gatekeeper/codesign cache) where all 4 attempts fail identically.

**Classification, not blanket retry**: on a completed `build-tests-*` run with failures, this fetches every failed job's log and checks it against the *exact same* transient-error pattern already used by the in-process retry (`TimeoutPolicy did not complete|Access token is unavailable|entitlement groups and 0 free entitlements|License activation has failed|No valid Unity Editor license found|License is not active`). Only if **every** failed job matches does it call `reRunWorkflowFailedJobs`, gated to run once (`run_attempt == 1`). Any failure that doesn't match — a real compile error, a genuine license misconfiguration, anything else — is left alone for a human or agent to look at.

Note: `workflow_run`-triggered workflows only activate from the copy on the default branch, so this is a standalone PR against `main` rather than bundled into #844.

## Test plan
- [x] `actionlint` — clean
- [ ] Will only be verifiable once merged and a real `build-tests-*` run fails with the classified pattern

## Summary by CodeRabbit

* **New Features**
  * Added automatic one-time retries for select build workflows when failures match a known transient Unity licensing issue.
  * Non-matching failures, log retrieval errors, and repeat attempts remain available for manual review.

### DIFF
--- a/.github/workflows/auto-retry-known-flakes.yml
+++ b/.github/workflows/auto-retry-known-flakes.yml
@@ -1,0 +1,92 @@
+name: Auto-retry known-transient CI flakes
+
+# Careful, narrow auto-healing for this repo's OWN test CI - not a blanket
+# "retry until green" mechanism. game-ci/cli's Unity build/activate steps
+# already retry a specific, documented set of transient licensing errors
+# in-process (up to --licenseRetryMaxAttempts, default 4, exponential
+# backoff - see dist/platforms/{mac,ubuntu}/steps/{activate,build}.sh). That
+# retries on the SAME runner instance, so it can't help when the underlying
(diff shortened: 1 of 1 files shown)


frostebite:
Consolidating into #844 (the single thin-wrapper PR) per request — this workflow is now included there instead of as a separate PR.

